PR #3, opened 2015-01-26
switches to using googles directory api

steppenwells:
as opposed to the data api which is being retired.


theefer:
Great! :+1: 

I assume we'll have both sets of properties in the config file for a while as people migrate?

steppenwells:
we will indeed, in fact they are there already.

PR #4, opened 2015-02-06
handle null google sends of getGroups method

lindseydew:
The Google API was throwing a null in cases where users did not have the correct groups.

This handles that case, so that the correct error messages are sent in cases where user is not Guardian, or has 2fa enabled.


steppenwells:
Looks good +1 etc

PR #5, opened 2015-02-06
Validation check update

steppenwells:
updates the play implementation to check user validity every request.

the old behaviour (only checking validity once per system per google session) can be turned back on by overriding the cacheValidation setting to true.

Note this is not implemented in the core library so the flex (scalatra) implementation will need to be updated to behave similarly.


sihil:
:+1: particularly the full stop in the docs.

PR #6, opened 2015-03-05
Extended expiry

jamesgorrie:
I've left it pretty simple for now hoping not to make a mess of it.

I also thought we could have this as a property that the apps using this could set, but the logic is always going to be "expire the API requests after the App" so I thought since all of that is accessible here, let's do it here.

I am not sure what needs to be changed in terms of versioning etc as I've never published scala stuff.

### DIFF
--- a/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/AuthenticatedUser.scala
+++ b/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/AuthenticatedUser.scala
@@ -5,5 +5,5 @@ import java.util.Date
 case class AuthenticatedUser(user: User, authenticatingSystem: String, authenticatedIn: Set[String], expires: Long, multiFactor: Boolean) {
 
   def isExpired = expires < new Date().getTime
-  def hasExpiryExtension(extension: Long) = (expires + extension) > new Date().getTime
+  def isInGracePeriod(period: Long) = (expires + period) > new Date().getTime
 }
--- a/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/AuthenticatedUser.scala
+++ b/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/AuthenticatedUser.scala
@@ -5,4 +5,5 @@ import java.util.Date
 case class AuthenticatedUser(user: User, authenticatingSystem: String, authenticatedIn: Set[String], expires: Long, multiFactor: Boolean) {
 
   def isExpired = expires < new Date().getTime
+  def hasExpiryExtension(extension: Long) = (expires + extension) < new Date().getTime
 }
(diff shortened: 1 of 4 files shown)


pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/AuthenticatedUser.scala
```diff
@@ -5,4 +5,5 @@ import java.util.Date
 case class AuthenticatedUser(user: User, authenticatingSystem: String, authenticatedIn: Set[String], expires: Long, multiFactor: Boolean) {
 
   def isExpired = expires < new Date().getTime
+  def hasExpiryExtension(extension: Long) = (expires + extension) > new Date().getTime
```
sihil: isExpiredWithGrace?

jamesgorrie: I prefer grace, thanks.
The important thing here though is to have a `def` that can find out if you are in that grace period, not that you've expired it to be used in [conjunction with isExpired](https://github.com/guardian/pan-domain-authentication/pull/6/files#diff-71a4f683684a360e5cf56449b5cbe2bdR194).

I can try and work that logic to work with another `is` I just couldn't think of one.

steppenwells: I'd second using grace throughout, extension make it sound like you're bumping the expiry date but grace / gracePeriod reflect what's going on.

isWithinExpiryGracePeriod could work as a method name.

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -37,6 +38,18 @@ trait AuthActions extends PanDomainAuth {
...
+  /**
+   * Adding an expiry extension to `APIAuthAction`s allows for a delay between an applications authentication and their
+   * respective API XHR calls expiring.
+   *
+   * This is particularly useful for SPAs where users have third party cookies disabled.
+   *
+   * @return the amount of delay between App and API expiry in milliseconds
+   */
+  def apiExpiryExtension: Long = 1.hour.toMillis
+
```
steppenwells: Do we want to default this to an hour? might be better to have no grace period by default and make apps override this to enable the grace period.

Either way the doc comment should mention what the default behaviour is as well as what the setting does.

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/AuthenticatedUser.scala
jamesgorrie: I think I just did that as soon as I read this - seems to read a lot nicer.

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala
jamesgorrie: Agreed again - leave it to the clients to decide their grace period...
This does however mean that if we're accessing Workflow and the Grid from Composer, one might fall out of grace and the other not.


I don't foresee this being a problem - probably a benefit for people to find their best / most secure grace period.

sihil:
This is looking good. As @steppenwells says I think it's worth documenting and explaining the reasoning behind it in the main README in addition to the javadocs.

:+1:  from me though.

PR #7, opened 2015-03-25
Refactor APIAuthAction to allow custom overriding of error results

theefer:
Currenty APIAuthAction returns error responses with an empty body. :-1: 

Different consumers may want to provide different error responses based on their API semantics, so this decouples the `Result`s of error responses from the extra logic needed for each case.


steppenwells:
Looks ok, but can you add something to the readme about how to customise the error responses

theefer:
I will do in a separate PR, the README needs a few more updates and I'd rather not make that PR include more unrelated stuff.

steppenwells:
ok, look forward to seeing it etc.

PR #9, opened 2015-03-27
Adds custom reads for UserInfo

kenoir:
Fix for the Google update to the userinfo response.

### DIFF
--- a/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/googleAuthModel.scala
+++ b/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/googleAuthModel.scala
@@ -26,7 +26,10 @@ object Token {
     (JsPath \ "id_token").read[String]
   )(Token.apply _)
 
-  def fromJson(json:JsValue):Token = Json.fromJson[Token](json).get
+  def fromJson(json:JsValue):Token = {
+    println(json)
+    Json.fromJson[Token](json).get
+  }
 }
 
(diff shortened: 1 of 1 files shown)


kenoir:
@steppenwells boop.

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/googleAuthModel.scala
```diff
@@ -26,7 +26,10 @@ object Token {
     (JsPath \ "id_token").read[String]
   )(Token.apply _)
 
-  def fromJson(json:JsValue):Token = Json.fromJson[Token](json).get
+  def fromJson(json:JsValue):Token = {
+    println(json)
```
steppenwells: remove

theefer:
You could use Boolean.toString

theefer:
Change to kind optionality presumably intentional? Looks good otherwise. :+1:

steppenwells:
Looks good :+1:

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/googleAuthModel.scala
```diff
@@ -26,7 +26,9 @@ object Token {
     (JsPath \ "id_token").read[String]
   )(Token.apply _)
 
-  def fromJson(json:JsValue):Token = Json.fromJson[Token](json).get
+  def fromJson(json:JsValue):Token = {
```
rtyley: Don't think anything really changes here does it? Can probably leave that change out.

theefer:
Nice fix :+1:

PR #10, opened 2015-04-29
Asymmetrical cookie signature

adamnfish:
This PR does two things:

1. replace the shared secret with a public/private keypair for signing and verifying logins
2. introduce a library for cookie verification, independent of the machinery used to fetch the secrets

Switching the shared secret for a public/private keypair means verifying existing logins does not need access to any secret information. The core library is designed to make it easy to fetch those secrets so they can be integrated in your webapp. The new verification library doesn't include any of this and just offers a function for doing the verification without bringing a wealth of ancillary library dependencies into people's projects. The core logic in the verification library is re-used in the core and play modules (we also add tests for this core logic).

I've updated the documentation to reflect the changes. You'll notice it refers to `"0.2.7-SNAPSHOT"` at the moment, because there is currently no published version of the verification library. This version number will need to be updated along with the others at publish time, as normal.

Unfortunately, this PR's diff is showing moved files as deletions rather than renames (you'll see the renames if you look commit-by-commit) which makes it look much larger than it is, my apologies.


adamnfish:
Note: I've added a public/private keypair into the `local.dev-gutools.co.uk` domain's settings for my own testing. Before merging this PR we'll need to make sure those properties are added to the other domains. Likely we can use one shared "dev" keypair for CODE, local etc similar to the way the secret is done now.

adamnfish:
After discussion with @steppenwells, we've established that the library should support both formats during a transition period. This will ensure we can prevent any interruption of service when the cookie format changes.

As such, further changes are required.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PanDomain.scala
```diff
@@ -0,0 +1,31 @@
...
+        NotAuthorized(authedUser)
+      }
+    } catch {
+      case e: Exception =>
+        InvalidCookie(e)
+    }
+  }
+
+  val guardianValidation: AuthenticatedUser => Boolean = { authedUser =>
+    (authedUser.user.emailDomain endsWith "@guardian.co.uk") && authedUser.multiFactor
```
ScalaWilliamTG: Should be
```scala
emailDomain == "guardian.co.uk"
```

As domains don't contain @

adamnfish: Great spot. It used to be `email endsWith "@guardian.co.uk"` so this is a mistake when switching it. It should have been tested, I'll fix that.

ScalaWilliamTG:
We'd like a released version of this pull request, as Preview changes in MAPI will be merged to master soon.

adamnfish:
This has been replaced by the Phase I pull request, #11

PR #11, opened 2015-05-14
Asymmetrical cookie signature - phase 1

adamnfish:
This pull request is the first phase in introducing asymmetrically signed cookies to the pan domain auth library. This phased change ensures there will be no interruption of service for users.

It builds on the changes in #10 - re-introducing support for the old cookie verification. The following diff shows how this PR differs from the initial proposal.
https://github.com/guardian/pan-domain-authentication/compare/asymmetrical-cookie-signature...asymmetrical-cookie-signature-phase-I

The required steps will be:

1. (this PR) introduce new asymmetrically signed cookie and drop it under a different name
2. perform the check against the new cookie but continue to drop the legacy cookie
3. deprecate the legacy cookie and use the new cookie value on both cookie names
4. remove the temporary new cookie

Each step will need a full deploy of every product, but taking this approach means there will be no interruption of service.


steppenwells:
Looks fine and approach should work.

:+1:

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PanDomain.scala
```diff
@@ -0,0 +1,53 @@
...
+      case e: Exception =>
+        InvalidCookie(e)
+    }
+  }
+
+  /**
+   * Cookie check that includes the legacy cookie. This is required during a period of transition
+   * between the old cookie and the new assymetric one.
+   */
+  def authStatusWithLegacyCheck(cookieData: String, publicKey: String, secret: String): AuthenticationStatus = {
```
lindseydew: Is it worth having a unit test to check that the different possibilities of the cookie value are parsed by the correct methods?

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
```diff
@@ -0,0 +1,64 @@
...
+            deserializeAuthenticatedUser(new String(Base64.decodeBase64(data)))
+          } else {
+            throw new CookieSignatureInvalidException
+          }
+        } catch {
+          case e: SignatureException =>
+            throw new CookieSignatureInvalidException
+        }
+      case _ => throw new CookieParseException
+    }
```
lindseydew: Not really a big deal for this PR, but kind of feel this method should return an Option[AuthenticatedUser]m, or some such value to avoid the nested try catch block in the authStatusWithLegacyCheck method

lindseydew:
Looks good so far :+1:  Let me know when there is a snapshot available and I'll test on some of our services.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
adamnfish: I couldn't agree more. This style is a nighmare and elsewhere in the code depends on the side effect of an exception being thrown when this is run. This is definitely a candidate for an overhaul but this is the existing logic simply moved to a new file so I've not attempted to extensively refactor everything.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PanDomain.scala
adamnfish: Yeah this probably is worth doing. I didn't duplicate all the tests for this temporary method but it's totally critical to the library's workings so I should probably just go ahead and do that for now.

adamnfish: I've added a couple of tests for this, great idea.

lindseydew:
Tested this locally with permissions and workflow, and the authentication mechanism appears to work ok. I can't test this in an environment because teamcity doesn't have access to the Snapshot repo. Suggest merging this, and we can do a more thorough regression test on release environments. @nealmadlani, @rcrphillips does this approach sound sensible to you?

lindseydew:
Currently if I don't supply pandomain access keys, S3 throws an Access Denied exception, which causes the authentication method to fail. Suspect this change doesn't address this dependency? If not, think it is worth releasing anyway, but would be good at some point to deprecate the requirement for specific aws keys.

sihil:
@lindseydew I'm pretty sure if `awsCredentials = None` in the `AuthActions` trait then it will fallback to the default provider chain. Check your instance profile permissions on the S3 bucket? I think this is how we use it composer-restorer.

sihil:
What is the storage strategy for the public and private keys here? Different buckets? Or folders within buckets?

lindseydew:
@sihil yeah that's a better way of handling the s3 reads. I will probably implement that anyway. @adamnfish mentioned this approach should remove the need for s3 reads entirely, but I this pass still relies on access to the bucket.

adamnfish:
This change introduces a new jar, the verification library, which doesn't
have any dependency on s3. The actual pan domain library still does and
this doesn't mean to change that? One of the reasons for using a separate
library is that the pan domain library's side effects (e.g. S3 reads) are
inexorably entwined. This is convenient for the library's normal use case,
but inflexible for others.

For the permission use case (among others), you should only need the
verification jar?

I'm at a conference at the moment so I can chat it over with you on Friday
when I'm back in the office properly if that helps? Otherwise drop me am
email, I'm checking them between talks!

On Wednesday, 3 June 2015, Lindsey Dew <notifications@github.com> wrote:

> @sihil <https://github.com/sihil> yeah that's a better way of handling
> the s3 reads. I will probably implement that anyway. @adamnfish
> <https://github.com/adamnfish> mentioned this approach should remove the
> need for s3 reads entirely, but I this pass still relies on access to the
> bucket.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/guardian/pan-domain-authentication/pull/11#issuecomment-108261822>
> .
>

-- 

------------------------------
Visit theguardian.com. On your mobile and tablet, download the Guardian 
iPhone and Android apps theguardian.com/guardianapp and our tablet editions 
theguardian.com/editions.  Save up to 57% by subscribing to the Guardian 
and Observer - choose the papers you want and get full digital access. 
 Visit subscribe.theguardian.com

This e-mail and all attachments are confidential and may also be 
privileged. If you are not the named recipient, please notify the sender 
and delete the e-mail and all attachments immediately. Do not disclose the 
contents to another person. You may not use the information for any 
purpose, or store, or copy, it in any way.  Guardian News & Media Limited 
is not liable for any computer viruses or other material transmitted with 
or as part of this e-mail. You should employ virus checking software.
 
Guardian News & Media Limited is a member of Guardian Media Group plc. Registered 
Office: PO Box 68164, Kings Place, 90 York Way, London, N1P 2AP.  Registered 
in England Number 908396

lindseydew:
Right yeah that makes sense, thank you. 

It's a little tricky for me to verify the s3 logic from this PR because the teamcity can't download the snapshot version, but for now I will take @sihil approach for deprecating the panda keys.

lindseydew:
I think this version breaks the re-auth logic in workflow. I haven't delved into why, but I can show you the behaviour when you're back in the office - hope the conference is interesting!

adamnfish:
Following our conversation the verification library now contains a mechanism for fetching the public key. I've also improved the documentation to describe how to use this new functionality.

adamnfish:
Following subsequent discussions, the library now provides a recommended way to fetch the public key.

I've also updated the out-of-date crypto, so we don't need to make another breaking change straight after this one.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
```diff
@@ -0,0 +1,143 @@
...
+   * @param ec     implicit execution context to use for fetching the key
+   */
+  def getPublicKey(domain: String)(implicit client: Http, ec: ExecutionContext): Future[String] = {
+    getPublicSettings(domain) flatMap extractPublicKey
+  }
+
+  // internal functions for fetching and parsing the responses
+  private def fetchSettings(domain: String)(implicit client: Http, ec: ExecutionContext): Future[Either[Throwable, String]] = {
+    val req = host("s3-eu-west-1.amazonaws.com").secure / bucketName / s"$domain.settings.public"
+    client(req OK as.String).either
```
lindseydew: Is the contents of the public key the same for all environments?

theefer: Presumably not given that the private key should be different?

lindseydew:
Changes look good :+1: 
For testing it might make sense to test this with one app (composer?) on release and make sure the tools still work ok. Then once there is a stable jar we can move all the apps one by one? cc @rcrphillips

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
adamnfish: As @theefer says, no. This function takes the domain as an argument because the public key varies by domain (i.e. the environment). The naming convention is currently `<domain>.settings`, I'm adding a corresponding public file at `<domain>.settings.public`.

adamnfish:
@lindseydew agreed. We can test on release with a snapshot at first and then publish the jar before we take it to PROD. At that point mobile can start to take advantage of it as well.

lindseydew:
I've put a version of this library on workflow CODE. 
@rcrphillips if you could sanity check this look ok, and auth still works on apps on CODE? 
then @adamnfish could you publish? and we can sort out migrating each of the apps over.

rcrphillips:
:+1: 
presence, assign to and no-sign-in all seems well

lindseydew:
Testing went well, @adamnfish are you happy for me to merge and publish?

adamnfish:
Great news!

Yes, if you've the chance then go ahead! We'll need to upload the keys to each environment as well. I can do it, or we can do it together to make sure we both know for the future? Is the plan to test this with the released jar on the release environment next?

lindseydew:
Had forgotten that step. Happy for either you to let me know when you've done it, or to do it together. Don't feel confident enough to do it by myself.

PR #13, opened 2015-07-15
Add a duplicate auth project for play 2.4.0

rich-nguyen:
This is an ugly solution to the problem where pan-domain clients may be using an incompatible version of Play from the expected version to be provided. For example, pan-domain will throw a runtime error if used with Play 2.4.0, because of a breaking signature change in the WS api.


rich-nguyen:
verbal :+1: from @steppenwells

rich-nguyen:
( @steppenwells would rather not type right now)

PR #14, opened 2015-07-28
Revert "Asymmetrical cookie signature - phase 1"

lindseydew:
Reverts guardian/pan-domain-authentication#11


adamnfish:
This was reverted because it didn't compile, @lindseydew realised that's because she's on Java 7. We're ok to go ahead with this now.

lindseydew:
Looks good :+1:

PR #17, opened 2015-08-12
Fix character encoding in new cookie

adamnfish:
We no longer rely on the platform's default character encoding being correct.

(There was a missing `getBytes("UTF-8")` call in the CookieUtils)


theefer:
:+1:

theefer:
How come I'm not having any issues if this hasn't been merged and released?

Asking out of interest, but will merge and release soon.

adamnfish:
Why hello. I confess I forgot all about this. Pretty sure you raised it from spotting an issue somewhere? I forget where though?

theefer:
Ah yes I thought it was my name on the Workflow page. Doesn't look wrong now, not sure if a case of using old/new cookies or what?

PR #18, opened 2015-10-19
Use credentialsProvider instead of awsCredentials

philmcmahon:
Otherwise when using temporary credentials the library will stop working after the credentials expire. This should fix that.

### DIFF
--- a/pan-domain-auth-example/app/controllers/ExampleAuthActions.scala
+++ b/pan-domain-auth-example/app/controllers/ExampleAuthActions.scala
@@ -1,6 +1,7 @@
 package controllers
 
-import com.amazonaws.auth.{BasicAWSCredentials, AWSCredentials}
+import com.amazonaws.auth.{AWSCredentialsProvider, BasicAWSCredentials}
+import com.amazonaws.internal.StaticCredentialsProvider
 import com.amazonaws.regions.Region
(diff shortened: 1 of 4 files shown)


pan-domain-auth-example/app/controllers/ExampleAuthActions.scala
```diff
@@ -25,7 +26,7 @@ trait ExampleAuthActions extends AuthActions {
   lazy val awsSecretAccessKey: String = config.getString("aws.secret").get
   lazy val awsKeyId: String = config.getString("aws.keyId").get
 
-  override lazy val awsCredentials: Option[AWSCredentials] = Some(new BasicAWSCredentials(awsKeyId, awsSecretAccessKey))
+  override lazy val awsCredentialsProvider: AWSCredentialsProvider = new StaticCredentialsProvider(new BasicAWSCredentials(awsKeyId, awsSecretAccessKey))
```
sihil: A comment here suggesting that a hard coded key should be a last resort?

sihil:
This looks good to me. :+1: 

An slight aside: flexible-content uses 0.2.6 so bumping to this release would also pick up the asymmetric cookie work too. Should be fine, just worth bearing in mind.

Issue #19, opened 2015-10-19
Static name of S3 bucket: pan-domain-auth-settings


vimvim:
S3 bucket names shared between all users in the AWS system so there is impossible to create another S3 bucket "pan-domain-auth-settings" in the another AWS accounts.

rrees:
That's true. Feel free to contribute a patch that parameterises the name so it is easy to use another bucket name.

piuccio:
seems like this issue can be closed, right?

PR #20, opened 2015-10-26
Parametrize Bucket Name with environment variable

m4tx:
Bug: #19


theefer:
:+1:

theefer:
Merged and released as 0.2.9, JAR should be publishing now.

theefer:
Thanks for the contribution!

PR #12, opened 2015-06-17
Add a section on cross-account access for an S3 bucket

sihil:

PR #21, opened 2015-11-06
Add assume role script

lindseydew:
Adds a template for accounts other than workflow to create a role for their account to access the panda bucket. Intention is to deprecate using the bucket policy to grant access to accounts outside workflow.

@seb please let me know if you want to test this out with the grid.

cc @sihil


jamespamplin:
Do we need some usage info in the readme?

panda.json
```diff
@@ -0,0 +1,73 @@
+{
+    "AWSTemplateFormatVersion": "2010-09-09",
+    "Description": "Facia tool template to create cross account roles.",
```
adamnfish: Is this specific to Facia?

lindseydew:
@jamespamplin that's a good thought I will add this in.

README.md
```diff
@@ -28,6 +28,23 @@ if they are valid then the cookie is updated to indicate this otherwise an error
...
+access to the pan-domain bucket in the cloudformation script.
+For applications outside the Workflow account there is an example sts:AssumeRole script. You need to add your AWS accountId in the 
+parameters and upload this to the Workflow account and get the roleArn.
+
+In your PanDomainAuthActions include this:
+
+```
+override def awsCredentialsProvider  = {
+  new AWSCredentialsProviderChain(new STSAssumeRoleSessionCredentialsProvider({{roleArn}}, "yourapp"),
+                                  new ProfileCredentialsProvider("workflow"))
```
theefer: Should that be `{{your profile in dev}}`?

theefer: Just saw the note below. Would be nice to have a way that doesn't require ad-hoc credentials for the workflow account.

lindseydew: Yeah it would be nice to have as few dependencies as possible for local dev, agreed. It is up to the client how they implement the provider chain, and so you could chose instead to use keys to default to if you'd prefer. I think in general it makes sense for applications that need access to resources in other accounts to use the credentials of the profile its accessing, which is why I've suggested this approach. Happy to discuss though?

sihil:
@lindseydew can this be merged? Think it supercedes #12?

lindseydew:
Yep - should be ok. Do you mind giving it a plus one?

sihil:
The new GH pull-request views are helping me catch up.

This now has a long overdue 👍

philwills:
*bump*

lindseydew:
Don't think the use case in relevant anymore.

PR #22, opened 2015-11-10
Throw GoogleAuthException on 'anti forgery token did not match' error

theefer:
Change the type of error thrown when encountering a mismatching anti-forgery token, to match the errors thrown [in other cases when parsing the Google response](https://github.com/guardian/pan-domain-authentication/blob/e3f11de0fdb02a510176abd10b2b3492f6db3fa4/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/GoogleAuth.scala#L32-L44), in order to make it easier to match.

Alternatively I could amend our code to match `IllegalArgumentException` (though would have to do string matching into the error message which sounds wrong), if people think this shouldn't be the same type of exception.

Thoughts?

This is part of the work to reduce auth-related errors in the Grid. The idea would be to return a 400 whenever encountering any `GoogleAuthException`, hence encouraging the client to optionally retry the auth flow.


theefer:
Ping @adamnfish @sihil ?

theefer:
cc @akash1810

adamnfish:
Yeah this looks reasonable and good to make it easier to handle expected errors. I don't love that the error is still represented just as a string (you're just moving the string matching logic that you don't want to do) but this is consistent.

PR #23, opened 2015-11-11
Retry 3 times when encountering Google flaky 500 error

theefer:
We don't know why Google sometimes returns a 500 error with a snippet of JSON helpfully describing it as "Backend Error", but I thought hey why not retry it a few times when that happens in case it's a race condition or whatnot.

Such errors are observed on all the apps using panda, according to our logging server.

Sadly don't really have a way to reproduce this, apart from running in PROD.

Happy for the Grid to be the guinea pig for this.

### DIFF
--- a/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/GoogleAuth.scala
+++ b/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/GoogleAuth.scala
@@ -2,8 +2,10 @@ package com.gu.pandomainauth.service
 
 import com.gu.pandomainauth.model.{User, GoogleAuthSettings, AuthenticatedUser}
 import play.api.mvc.Results.Redirect
-import play.api.mvc.{Call, Result, RequestHeader}
-import scala.concurrent.{ExecutionContext, Future}
+import play.api.mvc.{Result, RequestHeader}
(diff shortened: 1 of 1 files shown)


theefer:
cc @akash1810

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/GoogleAuth.scala
```diff
@@ -99,4 +93,40 @@ class GoogleAuth(config: GoogleAuthSettings, system: String, redirectUrl: String
...
     }
   }
-}
\ No newline at end of file
+
+
+  private def requestToken(tokenEndpoint: String, code: Seq[String], retries: Int = 3): Future[WSResponse] = {
+    WS.url(tokenEndpoint).post {
+      Map(
+        "code" -> code,
```
akash1810: is there a chance `code` will expire whilst in the retry loop, which will (hopefully) throw a different error than `Backend Error`.

theefer: I don't know, but if so the other error will just break the recursion and be passed to the caller?

akash1810:
:+1:

PR #26, opened 2016-04-22
Catch cookies parse exceptions

DiegoVazquezNanini:
This PR does a try catch while attempting to parse the cookie while we finish the transition to the new assymetric cookie.
@adamnfish do you want to review this PR ?

### DIFF
--- a/pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
+++ b/pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
@@ -117,10 +117,18 @@ trait AuthActions extends PanDomainAuth {
 
     // use legacy cookie in the check for now, even though both are dropped
     // phase two of the rollout will be to use the assymetric cookie
+    /**
+      * This is required during a period of transition       * between the old cookie and the new assymetric one.
+      */
     GoogleAuth.validatedUserIdentity(token).map { claimedAuth =>
       val authedUserData = existingCookie match {
         case Some(c) => {
-          val existingAuth = LegacyCookie.parseCookieData(c.value, settings.secret)
+          val existingAuth = try {
+            LegacyCookie.parseCookieData(c.value, settings.secret)
+          } catch {
+            case e: Exception =>
+              CookieUtils.parseCookieData(c.value, settings.publicKey)
+          }
           Logger.debug("user re-authed, merging auth data")
 
           claimedAuth.copy(
--- a/pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
+++ b/pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
@@ -118,7 +118,7 @@ trait AuthActions extends PanDomainAuth {
     // use legacy cookie in the check for now, even though both are dropped
     // phase two of the rollout will be to use the assymetric cookie
     /**
-      * This is required during a period of transition       * between the old cookie and the new assymetric one.
+      * This is required during a period of transition between the old cookie and the new assymetric one.
       */
     GoogleAuth.validatedUserIdentity(token).map { claimedAuth =>
       val authedUserData = existingCookie match {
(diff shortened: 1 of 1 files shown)


pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -117,10 +117,18 @@ trait AuthActions extends PanDomainAuth {
 
     // use legacy cookie in the check for now, even though both are dropped
     // phase two of the rollout will be to use the assymetric cookie
+    /**
```
adamnfish: You should replace the comment above with something along the lines of "it's now time to do both checks".

pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -117,10 +117,18 @@ trait AuthActions extends PanDomainAuth {
...
+      */
     GoogleAuth.validatedUserIdentity(token).map { claimedAuth =>
       val authedUserData = existingCookie match {
         case Some(c) => {
-          val existingAuth = LegacyCookie.parseCookieData(c.value, settings.secret)
+          val existingAuth = try {
+            LegacyCookie.parseCookieData(c.value, settings.secret)
+          } catch {
+            case e: Exception =>
+              CookieUtils.parseCookieData(c.value, settings.publicKey)
```
adamnfish: I'd consider doing this the other way round now - trying the modern cookie and falling back to the Legacy one where required. I doubt there are any services that don't drop the asymmetrical cookie by now.

adamnfish:
Good change! I had a couple of comments about the details but it's definitely time to switch Panda so it's reading the asymmetrical cookie by default.

DiegoVazquezNanini:
@sihil @steppenwells could please had a look at PR. Thanks

pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -115,12 +115,16 @@ trait AuthActions extends PanDomainAuth {
...
-    // phase two of the rollout will be to use the assymetric cookie
+    // use both parsers until we complete the transition to only set the assymetric cookie.
     GoogleAuth.validatedUserIdentity(token).map { claimedAuth =>
       val authedUserData = existingCookie match {
         case Some(c) => {
-          val existingAuth = LegacyCookie.parseCookieData(c.value, settings.secret)
+          val existingAuth = try {
+            CookieUtils.parseCookieData(c.value, settings.publicKey)
+          } catch {
+            case e: Exception =>
```
mchv: I think you should use `scala.util.control.NonFatal` instead of `Exception`

DiegoVazquezNanini: I am interested in only catching the exception thrown by parseCookieData
so I am confident that it will do the right thing.
Thanks, now I know about NonFatal

rrees:
👍

pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
sihil: I agree with @mchv, that this should be `NonFatal` - otherwise you might wind up catching internal JVM exceptions that should bubble up to the top of the application.

steppenwells: Or alternatively just catch the specific exception that is thrown that you care about. The exception will be one of CookieSignatureInvalidException or CookieParseException found in:

https://github.com/guardian/pan-domain-authentication/blob/master/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/model/exceptions.scala

DiegoVazquezNanini:
Thanks for the reviews I plan to release it asap.

PR #27, opened 2016-05-03
Specify dispatch-core library to resolve the incompatibilies issue

jennysivapalan:
Play 2.4 has incompatibilities with Ning and dispatch libraries. This is problematic for users of PublicSettings (e.g https://github.com/guardian/login.gutools). This PR specifies the dispatch-core library which will resolve the issue.

See here for more details: https://github.com/AnormCypher/AnormCypher/issues/44

I've tested this change locally for the specific use case that login.gutools requires. @steppenwells @adamnfish please could you help me identify what other use cases I should test?

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -17,7 +17,8 @@ object Dependencies {
   val playLibs_2_4_0 = Seq(
     "com.typesafe.play" %% "play" % "2.4.0" % "provided",
     "com.typesafe.play" %% "play-ws" % "2.4.0" % "provided",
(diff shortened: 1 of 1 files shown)


adamnfish:
This looks like a sensible change to me, good catch. Probably worth testing in CODE, but I'm sure it'll be fine?

jennysivapalan:
I chatted to Ed Tools team at standup. The plan is to make a release of this and everyone will test their apps against the new library version. Any issues I'll raise a fix.

philmcmahon:
:+1:

PR #28, opened 2016-06-07
Add Play 2.5 version of library

sihil:
I'm converting restorer to Play 2.5 so we need a compatible version of the library that uses DI for obtaining the `WSClient`. This has meant modifying the trait such that the client is provided there.

I've noted https://github.com/guardian/play-googleauth/pull/39 from @adamnfish that should simplify everything - but that's a little too much yak shaving for now.

Best reviewed by looking at https://github.com/guardian/pan-domain-authentication/pull/28/commits/fb05d7bc514452ca0a6d696f4fff7194584d2804


sihil:
I've proven this on restorer using a locally published snapshot.

sihil:
I've redone this using two commits - worth reviewing the second commit as this makes it clear what the changes are from Play 2.4. The first commit is just the copying of the files from Play 2.4 into a new Play 2.5 module.

piuccio:
👍  for what it's worth

PR #29, opened 2016-06-07
Remove 2.10 cross build

sihil:
Play 2.5 abandons Scala 2.10 and it turns out that it is impossible to configure cross builds differently for different sub-projects. As a result, and given that it is 2016 after all, I am making the executive decision to abandon Scala 2.10 and whilst I'm at it Play 2.3. 

More yak shaving FTW.


pan-domain-auth-example/app/controllers/AdminController.scala
```diff
@@ -1,19 +1,24 @@
...
+
+import play.api.Configuration
+import play.api.mvc.{Action, Controller}
 import play.api.Play.configuration
 import play.api.Play.current
+import play.api.libs.ws.WSClient
 
 
-object AdminController extends Controller with ExampleAuthActions {
+class AdminController @Inject() (override val config: Configuration, override val wsClient: WSClient)
```
piuccio: I've heard people arguing that compile time DI is better than run time.

But anyway, this file is just an example...

jennysivapalan:
👍

pan-domain-auth-example/app/controllers/AdminController.scala
sihil: Yep - I finally bit the bullet and worked out how to do compile time DI - the flexible-restorer project is a beautiful example of this. However, it adds two more classes and I didn't think it was worth the hassle in this case. Having said that it would be pretty easy.

LATaylor-guardian: I'd argue that it's worth converting to compile time dependency injection now at the earliest and least painful opportunity. Especially since you've got over the hump of the learning curve already. Assuming this project is going to continue to be developed on, the use of runtime dependency injection will only continue to grow. It's unlikely, but developers may also look as this project as a template in the future and decide to use runtime dependency injection as they have a working example. 

All in all less magic and craziness. That's my two cents. 😄

PR #30, opened 2016-06-17
Read the asymmetric cookie rather than the legacy cookie

sihil:
@piuccio pointed out yesterday that we are still entirely relying on the Legacy cookie in Pan Domain authed Play apps. Turns out that there also seem to be chunks of code that simply shouldn't work or don't do what we would expect.

There are two cookies - the legacy one with the name from `PublicSettings.cookieName` - which must be decoded with the `LegacyCookie` and the `secret`. Secondly there is the new one with the name `PublicSettings.assymCookieName` (if it wasn't going to break everything I'd fix this spelling) which must be decoded with `CookieUtils` and the `publicKey`.

There are a number of places where we try to decode a cookie with either the wrong decoder or the wrong key (the latter throws an exception, so some paths are clearly never being used). Either way - we should now be in a position where we solely rely on the asymmetric cookie, so we should never read from `PublicSettings.cookieName` or use `LegacyCookie` or `secret` for decoding it.

It also turns out that in some places we were discarding decoded cookies entirely - despite comments to the contrary - I've tried to fix those.

### DIFF
--- a/pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
+++ b/pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
@@ -198,7 +198,7 @@ trait AuthActions extends PanDomainAuth {
    */
   def extractAuth(request: RequestHeader): AuthenticationStatus = {
     readCookie(request).map { cookie =>
-      PanDomain.authStatusWithLegacyCheck(cookie.value, settings.publicKey, settings.secret) match {
+      PanDomain.authStatus(cookie.value, settings.publicKey) match {
         case Expired(authedUser) if authedUser.isInGracePeriod(apiGracePeriod) =>
           GracePeriod(authedUser)
         case authStatus @ Authenticated(authedUser) =>
--- a/pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
+++ b/pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
@@ -119,17 +119,12 @@ trait AuthActions extends PanDomainAuth {
     GoogleAuth.validatedUserIdentity(token).map { claimedAuth =>
       val authedUserData = existingCookie match {
         case Some(c) => {
-          val existingAuth = try {
-            CookieUtils.parseCookieData(c.value, settings.publicKey)
-          } catch {
-            case NonFatal(e) =>
-              LegacyCookie.parseCookieData(c.value, settings.secret)
-          }
+          val existingAuth = CookieUtils.parseCookieData(c.value, settings.publicKey)
           Logger.debug("user re-authed, merging auth data")
 
           claimedAuth.copy(
             authenticatingSystem = system,
-            authenticatedIn = Set(system),
+            authenticatedIn = existingAuth.authenticatedIn ++ Set(system),
             multiFactor = checkMultifactor(claimedAuth)
           )
         }
@@ -154,11 +149,11 @@ trait AuthActions extends PanDomainAuth {
 
 
   def readAuthenticatedUser(request: RequestHeader): Option[AuthenticatedUser] = readCookie(request) map { cookie =>
-    CookieUtils.parseCookieData(cookie.value, settings.secret)
+    CookieUtils.parseCookieData(cookie.value, settings.publicKey)
   }
 
 
-  def readCookie(request: RequestHeader): Option[Cookie] = request.cookies.get(PublicSettings.cookieName)
+  def readCookie(request: RequestHeader): Option[Cookie] = request.cookies.get(PublicSettings.assymCookieName)
 
   def generateCookies(authedUser: AuthenticatedUser): List[Cookie] = List(
     Cookie(
--- a/pan-domain-auth-play_2-5/src/main/scala/com/gu/pandomainauth/action/Actions.scala
+++ b/pan-domain-auth-play_2-5/src/main/scala/com/gu/pandomainauth/action/Actions.scala
@@ -122,21 +122,15 @@ trait AuthActions extends PanDomainAuth {
 
     val existingCookie = readCookie(request) // will be populated if this was a re-auth for expired login
 
-    // use both parsers until we complete the transition to only set the assymetric cookie.
     GoogleAuth.validatedUserIdentity(token)(request, global, wsClient).map { claimedAuth =>
       val authedUserData = existingCookie match {
         case Some(c) => {
-          val existingAuth = try {
-            CookieUtils.parseCookieData(c.value, settings.publicKey)
-          } catch {
-            case NonFatal(e) =>
-              LegacyCookie.parseCookieData(c.value, settings.secret)
-          }
+          val existingAuth = CookieUtils.parseCookieData(c.value, settings.publicKey)
           Logger.debug("user re-authed, merging auth data")
 
           claimedAuth.copy(
             authenticatingSystem = system,
-            authenticatedIn = Set(system),
+            authenticatedIn = existingAuth.authenticatedIn ++ Set(system),
             multiFactor = checkMultifactor(claimedAuth)
           )
         }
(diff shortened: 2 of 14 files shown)


pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -119,17 +119,12 @@ trait AuthActions extends PanDomainAuth {
     GoogleAuth.validatedUserIdentity(token).map { claimedAuth =>
       val authedUserData = existingCookie match {
         case Some(c) => {
-          val existingAuth = try {
-            CookieUtils.parseCookieData(c.value, settings.publicKey)
```
sihil: This would never have worked as the cookie that is read is the legacy cookie.

pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -119,17 +119,12 @@ trait AuthActions extends PanDomainAuth {
...
-            case NonFatal(e) =>
-              LegacyCookie.parseCookieData(c.value, settings.secret)
-          }
+          val existingAuth = CookieUtils.parseCookieData(c.value, settings.publicKey)
           Logger.debug("user re-authed, merging auth data")
 
           claimedAuth.copy(
             authenticatingSystem = system,
-            authenticatedIn = Set(system),
+            authenticatedIn = existingAuth.authenticatedIn ++ Set(system),
```
sihil: Here, we merge in the existing systems - which should reduce the number of cookie re-writes between systems.

pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -154,11 +149,11 @@ trait AuthActions extends PanDomainAuth {
 
 
   def readAuthenticatedUser(request: RequestHeader): Option[AuthenticatedUser] = readCookie(request) map { cookie =>
-    CookieUtils.parseCookieData(cookie.value, settings.secret)
+    CookieUtils.parseCookieData(cookie.value, settings.publicKey)
   }
 
 
-  def readCookie(request: RequestHeader): Option[Cookie] = request.cookies.get(PublicSettings.cookieName)
+  def readCookie(request: RequestHeader): Option[Cookie] = request.cookies.get(PublicSettings.assymCookieName)
```
sihil: Read the new cookie here

pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -154,11 +149,11 @@ trait AuthActions extends PanDomainAuth {
 
 
   def readAuthenticatedUser(request: RequestHeader): Option[AuthenticatedUser] = readCookie(request) map { cookie =>
-    CookieUtils.parseCookieData(cookie.value, settings.secret)
```
sihil: This would never have worked as it is trying to read the new cookie using the legacy secret.

adamnfish:
Maybe deprecate `PublicSettings.cookieName` and `LegacyCookie.parse/generate` if you aren't deleting them now? Helps make sure they aren't being used anywhere.

pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
adamnfish: I think @DiegoVazquezNanini must have had a service that set the new format cookie to the old cookie name when he did this? That's a bit mad so removing it makes sense for the library as a whole but be aware @DiegoVazquezNanini this will probably be a breaking change.

pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
adamnfish: Great spot.

pan-domain-auth-play_2-5/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -122,21 +122,15 @@ trait AuthActions extends PanDomainAuth {
 
     val existingCookie = readCookie(request) // will be populated if this was a re-auth for expired login
 
-    // use both parsers until we complete the transition to only set the assymetric cookie.
     GoogleAuth.validatedUserIdentity(token)(request, global, wsClient).map { claimedAuth =>
       val authedUserData = existingCookie match {
         case Some(c) => {
-          val existingAuth = try {
-            CookieUtils.parseCookieData(c.value, settings.publicKey)
-          } catch {
```
adamnfish: Why are you taking out the exception handling? That'll still be required.

adamnfish: Oh ignore me, because this is just to fall back.

adamnfish:
If you remove the old cookie completely it might be worth considering a 'minor' version bump (instead of 'patch')

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/package.scala
```diff
@@ -0,0 +1,7 @@
+package com.gu
+
+package object pandomainauth {
+  case class PublicKey(key: String)
```
adamnfish: If you're just using a type as a wrapper like this have it `extend AnyVal` so it gets unpacked at runtime.

adamnfish:
Officially a breaking change now (apps need to provide a `PrivateKey` to the generate method instead of a string), so definitely worth a minor bump.

piuccio:
breaking changes should bump the [major](http://semver.org/).

sihil:
Not sure it's really a breaking change. Most implementations get their private key from the settings object which has been changed to return the appropriate type. Iff people are creating or parsing their own keys is it a breaking change.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
```diff
@@ -123,8 +123,8 @@ object PublicSettings {
       case Left(err) => Future.failed(err)
     }
   }
-  private[pandomainauth] def validateKey(pubKey: String): Either[Throwable, String] = {
-    if ("[a-zA-Z0-9+/\n]+={0,3}".r.pattern.matcher(pubKey).matches) Right(pubKey)
+  private[pandomainauth] def validateKey(pubKey: String): Either[Throwable, PublicKey] = {
```
adamnfish: Any reason this doesn't take `PublickKey` for consistency?

pan-domain-auth-verification/src/test/scala/com/gu/pandomainauth/PublicSettingsTest.scala
```diff
@@ -23,13 +23,13 @@ class PublicSettingsTest extends FreeSpec with Matchers with EitherValues with S
 
     "returns the key if it is valid" in {
       val key = TestKeys.testPublicKey
-      PublicSettings.validateKey(key).right.value shouldEqual key
+      PublicSettings.validateKey(key.key).right.value shouldEqual key
```
adamnfish: Why not put the `.key` on `val key = TestKeys.testPublicKey` to make this a little less "keyn"? (Or have the method take a `PublicKey` which would make more sense I think)

adamnfish:
:+1: though, nice one!

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
sihil: Couldn't make it work, come and have a look after lunch.

piuccio:
indeed it breaks facia tool / story packages, `1.0.0` would have been more semantic, but whatever...

PR #32, opened 2016-07-20
Keypair script

sihil:
A script to make it less error prone to generate a new pair of credentials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,10 @@ You can generate an rsa key pair as follows:
     openssl genrsa -out private_key.pem 4096
     openssl rsa -pubout -in private_key.pem -out public_key.pem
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -199,6 +199,10 @@ You can generate an rsa key pair as follows:
     openssl genrsa -out private_key.pem 4096
     openssl rsa -pubout -in private_key.pem -out public_key.pem
 
+There is a helper script in the root of this project that uses the commands about and outputs a new keypair in the format used by the panda settings file:
```
piuccio: isn't it maybe `above` instead of `about`?

piuccio:
👍  I was hoping this would supersede #21 which was never merged because of the grid, but I understand they are different things

PR #33, opened 2016-12-28
Checks if user is in specified groups

marialivia16:


jennysivapalan:
LGTM 👍

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/service/Google2FAGroupChecker.scala
```diff
@@ -55,7 +55,7 @@ class GoogleGroupChecker(config: Google2FAGroupSettings, bucket: S3Bucket) exten
 
   def checkGroups(authenticatedUser: AuthenticatedUser, groupIds: List[String]): Boolean = {
     val query = directory.groups().list().setUserKey(authenticatedUser.user.email)
-    if(groupIds.isEmpty) false else groupIds.foldLeft(true){(acc, groupId) => acc & hasGroup(query, groupId)}
```
jennysivapalan: Maybe worth logging the groupIds is empty?

marialivia16: Good idea. I wasn't sure what happens when we don't specify any groups. 

If the groupIds is empty we could also assume that we have no restrictions for the groups and this should return `true`.

marialivia16:
@jennysivapalan I made the change we discussed, are you happy to merge this? 👀

jennysivapalan:
👍

PR #34, opened 2017-01-19
Fix a couple of problems with the user validation logic.

adamnfish:
Moves auth validation logic into one place and adds tests to address the following issues:

1. it was impossible to override the `guardianValidation` check
2. users in the grace period skipped the validation checks

cc @DiegoVazquezNanini @jennysivapalan 

## 1
In the first case, for any projects using panda that wanted to enforce non-Guardian rules (for example changing the email domain or dropping the 2FA requirement), it was impossible to properly override the validateUser check. The overridable validateUser function was used at session creation time, but the `guardianValidation` function was erroneously hard-coded for the subsequent session validation. This caused very confusing results.

We now consistently apply the project-provided validateUser function. This may cause problems for existing applications but only in surprising cases and if the projects were already doing crazy things.

## 2
This was a potential security issue, although exploiting it would depend on finding a project that behaves in a way we'd probably class as "not recommended".

If the project-supplied validateUser function is impure then it's possible that it would being to fail at some point after the original session has been issued. For example, there might be a session-revocation feature that returns false if the user has been blacklisted in some way. This check would have been bypassed during the grace period providing an attacker with a window of opportunity to exploit a compromised session.

It's worth noting that while it's unlikely this affects any of our projects, the pan-domain-auth library was designed to support these use cases (this is why the `cache validation` feature exists). Ignoring questions about whether these features make any sense the way they are implemented in panda, this Pull Request fixes this edge case.

### DIFF
--- a/pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
+++ b/pan-domain-auth-play_2-4-0/src/main/scala/com/gu/pandomainauth/action/Actions.scala
@@ -25,7 +25,7 @@ trait AuthActions extends PanDomainAuth {
    * @param authedUser
    * @return true if the user is valid in your app
    */
-  def validateUser(authedUser: AuthenticatedUser): Boolean
+  def validateUser(authedUser: AuthenticatedUser): Boolean = PanDomain.guardianValidation(authedUser)
 
 
   /**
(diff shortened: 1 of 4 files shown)


jennysivapalan:
I've looked through the code and am definitely in agreement in point 1. I think point 2 and the code for it is a bit beyond my knowledge. I think in particular to code for `authStatus` is quite different (though I think a bit is inlining). 
Have we tested the library as a snapshot on one of the Guardian projects using it? If not I think we should do that and then I'd be happy to give this a 👍

mbarton:
I'll pick this up again and port across to the Play 2.6 implementation. I fixed the hard-coded `guardianValidation` problem in #45 without realising that it then meant we were not correctly respecting the `cacheValidation` setting any more

PR #35, opened 2017-07-20
Support Play 2.6

kelvin-chappell:
A new artefact for Play 2.6.

Tested with Campaign Central.

/cc @guardian/digital-cms

### DIFF
--- a/pan-domain-auth-play_2-6/src/main/scala/com/gu/pandomainauth/action/Actions.scala
+++ b/pan-domain-auth-play_2-6/src/main/scala/com/gu/pandomainauth/action/Actions.scala
@@ -1,0 +1,336 @@
+package com.gu.pandomainauth.action
+
+import com.gu.pandomainauth.model._
+import com.gu.pandomainauth.service._
+import com.gu.pandomainauth.{PanDomain, PanDomainAuth, PublicSettings}
+import play.api.Logger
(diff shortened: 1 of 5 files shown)


pan-domain-auth-play_2-6/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -0,0 +1,336 @@
...
+
+  /**
+    * Action that ensures the user is logged in and validated.
+    *
+    * This action is for page load type requests where it is possible to send the user for auth
+    * and for them to interact with the auth provider. For API / XHR type requests use the APIAuthAction
+    *
+    * if the user is not authed or the auth has expired they are sent for authentication
+    */
+  object AuthAction extends ActionBuilder[UserRequest, AnyContent] {
```
mbarton: Is the only that that has changed is the `ActionBuilder` syntax?

I'm wondering if there's a way of avoiding so much copy-paste in the code between 2.5 and 2.6, especially since it's security-related stuff.

kelvin-chappell: The only difference is the signature here and the addition of a parser and execution context.
So I'll have a look at reducing the duplication.

PR #36, opened 2017-11-10
fix play26 project compile

akash1810:
there were two compile errors:
- `method withHeaders in trait WSRequest is deprecated: Use withHttpHeaders or addHttpHeaders`
- `[error] Class org.joda.convert.FromString not found - continuing with a stub.`


philmcmahon:
👍

PR #39, opened 2018-01-30
Migrate to scala 2.12 and play 2.6

alexduf:
- remove play 2.4 and play 2.5 support as they aren't 2.12 compatible
- refactor the AuthAction so that they are play2.6 friendly
- use the provided execution context everywhere instead of using scala's
- remove dispatch (no scala 2.12) replace with okhttp


pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuth.scala
```diff
@@ -17,7 +17,7 @@ import scala.concurrent.duration.FiniteDuration
 
 trait PanDomainAuth {
 
-  lazy val actorSystem = ActorSystem()
+  def actorSystem: ActorSystem
```
alexduf: I couldn't find a good reason for PanDomain to use its own actor system

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
```diff
@@ -96,21 +96,29 @@ object PublicSettings {
...
-  def getPublicKey(domain: String)(implicit client: Http, ec: ExecutionContext): Future[PublicKey] = {
+  def getPublicKey(domain: String)(implicit client: OkHttpClient, ec: ExecutionContext): Future[PublicKey] = {
     getPublicSettings(domain) flatMap extractPublicKey
   }
 
   // internal functions for fetching and parsing the responses
-  private def fetchSettings(domain: String)(implicit client: Http, ec: ExecutionContext): Future[Either[Throwable, String]] = {
-    val req = host("s3-eu-west-1.amazonaws.com").secure / bucketName / s"$domain.settings.public"
-    client(req OK as.String).either
+  private def fetchSettings(domain: String)(implicit client: OkHttpClient, ec: ExecutionContext): Future[Either[Throwable, String]] = {
```
alexduf: This is the only significant change

NathanielBennett:
👍 @alexduf this looks OK to me.

alexduf:
Thanks! I'll release a major version bump as I'm breaking the compatibility

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuth.scala
alexduf: Now I have my answer. This was creating a new actor system for each controller, with one actor in it. A bit wasteful so there's a PR here to attempt to solve it: https://github.com/guardian/pan-domain-authentication/pull/41

PR #41, opened 2018-02-19
Fix pan domain settings refresher

alexduf:
The actor system was instantiated for each controller. This changed when
the actor system was passed down as a parameter, which in turn created
more problems as the domain settings refresher was instantiated for
each controller leading to a name collision.

This attempts to fix it by extracting the refresher logic out of the
trait, and requiring it to be injected in each controller

cc @regiskuckaertz

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,26 +244,15 @@ import com.gu.pandomainauth.model.AuthenticatedUser
 
 trait PanDomainAuthActions extends AuthActions {
 
-  import play.api.Play.current
-  lazy val config = play.api.Play.configuration
-
   override def validateUser(authedUser: AuthenticatedUser): Boolean = {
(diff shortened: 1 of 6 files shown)


regiskuckaertz:
it looks good to me 👍 I just wonder if it would be cleaner if the actor was created in a method that receives an Actor System, with another one to kill it on exit for good measure?

regiskuckaertz:
Scratch that, the sole purpose it to create that actor so no need to delay passing the system in.

alexduf:
Cool, also passing the actor system managed by Play means when the app is turned off by play (including in dev mode) the actor system will be brought down the right way.

@regiskuckaertz do you want to do a release local and test it before I merge and release?

regiskuckaertz:
Yep, doing it right now

regiskuckaertz:
All good 🚢

PR #42, opened 2018-02-20
Non blocking s3 call and further improvements

alexduf:
I was bothered by some more stuff: Calls to s3 are blocking, and it's still using the deprecated agents.

 - Don't block when on S3, my approach may not be deemed "scala" enough, please be candid while reviewing. It's a cheap way to avoid blocking on Play's actor system
 - Remove Agents, and update the actor library to the same as play 2.6.11
 - Don't use the global execution context as we have one provided by the actor system

cc @regiskuckaertz

### DIFF
--- a/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
+++ b/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
@@ -1,13 +1,14 @@
 package com.gu.pandomainauth
 
-import com.amazonaws.auth.{DefaultAWSCredentialsProviderChain, AWSCredentialsProvider}
-import com.amazonaws.regions.{Regions, Region}
+import java.util.concurrent.atomic.AtomicReference
+
+import com.amazonaws.auth.{AWSCredentialsProvider, DefaultAWSCredentialsProviderChain}
+import com.amazonaws.regions.{Region, Regions}
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 import scala.language.postfixOps
-import akka.actor.{Props, Actor, ActorSystem}
-import akka.agent.Agent
+import akka.actor.{Actor, ActorSystem, Props}
 import akka.event.Logging
 import com.gu.pandomainauth.model.PanDomainAuthSettings
 import com.gu.pandomainauth.service.{ProxyConfiguration, S3Bucket}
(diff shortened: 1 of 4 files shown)


pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
```diff
@@ -32,45 +36,34 @@ class PanDomainAuthSettingsRefresher(
   awsRegion: Option[Region] = Option(Region getRegion Regions.EU_WEST_1),
   proxyConfiguration: Option[ProxyConfiguration] = None
 ) {
+  private implicit val ec: ExecutionContext = actorSystem.dispatcher
   lazy val bucket = new S3Bucket(awsCredentialsProvider, awsRegion, proxyConfiguration)
 
-  private lazy val settingsMap = bucket.readDomainSettings(domain)
-  private lazy val authSettings: Agent[PanDomainAuthSettings] = Agent(PanDomainAuthSettings(settingsMap))
+  private lazy val settingsMap = bucket.readDomainSettingsBlocking(domain)
```
alexduf: Blocking during setup as we can't proceed without the initial value of the settings

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
```diff
@@ -32,45 +36,34 @@ class PanDomainAuthSettingsRefresher(
...
-  }
-
-  override def postRestart(reason: Throwable) {
-    reschedule
-  }
-
-  def reschedule {
-    context.system.scheduler.scheduleOnce(frequency, self, Refresh)
+    case Refresh =>
+      bucket.readDomainSettingsAsync(domain)
```
alexduf: But non blocking while updating the settings

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/service/S3Bucket.scala
```diff
@@ -1,29 +1,52 @@
...
+        props.load(domainSecretFile.getObjectContent)
+        props.asScala.toMap
+      }
+      promise.complete(result)
+    }
 
-    val domainSecretFile = s3Client.getObject(new GetObjectRequest(bucketName, domain + ".settings"))
-    val props = new Properties()
+    // The java way to span a new thread in order to block elsewhere than on Play's actor system
+    val thread = new Thread(() => fetch)
```
alexduf: That's where I would like feedback. It's cheap but really not ideal

regiskuckaertz: Can you use this [`TransferManager`](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/index.html?com/amazonaws/services/s3/transfer/TransferManager.html) class? It looks like this is what gives you asynchronous powers with S3 ⚡️

alexduf: Had never heard of that, it looks useful. Unfortunately there's no method to download to a buffer. It looks like it's only writing to a `File` and must be optimised for large file workflows.

PR #43, opened 2018-04-05
Add PR template to include Editorial Tools Team in PR by default.

susiecoleman:
Adding a template to PRs so that the editorial tools team will be included by default. We rely heavily on these libraries for our log in and this will make it easier for the team to keep track of changes that may affect our tools.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,3 @@
+
+@guardian/editorial-tools-sshaccess
```
akash1810: why ssh access instead of `@guardian/digital-cms`?

susiecoleman: Because I forgot digital-cms exists. Just did a search for `tools`

PR #44, opened 2018-04-06
Add google group checker

susiecoleman:
Adding a google group checker to `AuthActions`

@guardian/digital-cms


mbarton:
Was this removed when we switched to Play 2.6?

susiecoleman:
@mbarton it wasn't removed but `settings` was made private so it's no longer possible to construct a group checker in an app in the same way

mbarton:
Hmm, perhaps we should fix that back to how it was. Does this change allow you to configure the groups it is checking?

susiecoleman:
@mbarton yes. The as the parameters for checking come from a parameterless function that is instantiated when using the `AuthActions` trait. It is also still possible to write essentially the same checker in the fronts code but it's messier than it was previously and I thought given we have a multifactor checker in there the group checker could go in too

PR #45, opened 2018-12-04
Support other OAuth providers

mbarton:
Adds the ability to configure the OAuth discovery document and therefore use Panda with OAuth providers that are not google.

We were unfortunately hard-coding the `validateUser` function when [checking the auth status](https://github.com/guardian/pan-domain-authentication/blob/68db57f487073af2bf5f9081926873cc1b8aef83/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PanDomain.scala). This is now customisable correctly. 

The rationale for the change is to allow non-Guardian developers to set up and evaluate the [Grid](http://github.com/guardian/grid) using a [Cognito user pool](https://docs.aws.amazon.com/cognito/latest/developerguide/getting-started-with-cognito-user-pools.html) rather than setting up a Google API project. In the future they can use Cognito to delegate to any auth provider using SAML.

I've also added an example Cognito user pool CloudFormation template.

This is a breaking change:

- The calling code must now specify the Panda bucket name, `AWSCredentialsProvider`, `region` directly
   - This avoids other users defaulting to the Guardian bucket
   - It was possible to use an environment variable to customise it. This is now the responsibility of the calling code
   - This avoids us including static credentials in the examples since we no longer recommend them at the Guardian

- Rename `googleId` and `googleClientSecret` to `clientId` and `clientSecret`

- The name of the cookies must be set in configuration. For example:
  - `legacyCookieName`: `gutoolsAuth`
  - `assymCookieName`: `gutoolsAuth-assym`

- `discoveryDocumentUrl` must be set in configuration
  - Google: `https://accounts.google.com/.well-known/openid-configuration`
  - Example Cognito: `https://cognito-idp.eu-west-1.amazonaws.com/eu-west-1_nW3FKqRh0/.well-known/openid-configuration`

### DIFF
--- a/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
+++ b/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
@@ -19,20 +19,22 @@ import scala.concurrent.duration.FiniteDuration
   *
   * @param domain the domain you are authin agains
   * @param system the identifier for your app, typically the same as the subdomain your app runs on
+  * @param bucketName the bucket where the settings are stored
   * @param actorSystem the actor system to create the refresh actor
-  * @param awsCredentialsProvider optional credential provider
-  * @param awsRegion optional region
+  * @param awsCredentialsProvider AWS credential provider
+  * @param awsRegion AWS region
   * @param proxyConfiguration optional proxy configuration
   */
 class PanDomainAuthSettingsRefresher(
   val domain: String,
   val system: String,
+  bucketName: String,
   actorSystem: ActorSystem,
-  awsCredentialsProvider: AWSCredentialsProvider = new DefaultAWSCredentialsProviderChain(),
-  awsRegion: Option[Region] = Option(Region getRegion Regions.EU_WEST_1),
+  awsCredentialsProvider: AWSCredentialsProvider,
+  awsRegion: String,
   proxyConfiguration: Option[ProxyConfiguration] = None
 ) {
-  lazy val bucket = new S3Bucket(awsCredentialsProvider, awsRegion, proxyConfiguration)
+  lazy val bucket = new S3Bucket(bucketName, awsCredentialsProvider, awsRegion, proxyConfiguration)
 
   private lazy val settingsMap = bucket.readDomainSettings(domain)
   private lazy val authSettings: Agent[PanDomainAuthSettings] = Agent(PanDomainAuthSettings(settingsMap))
--- a/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
+++ b/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
@@ -31,7 +31,7 @@ class PanDomainAuthSettingsRefresher(
   bucketName: String,
   actorSystem: ActorSystem,
   awsCredentialsProvider: AWSCredentialsProvider,
-  awsRegion: String,
+  awsRegion: Regions,
   proxyConfiguration: Option[ProxyConfiguration] = None
 ) {
   lazy val bucket = new S3Bucket(bucketName, awsCredentialsProvider, awsRegion, proxyConfiguration)
--- a/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/PanDomainAuthSettings.scala
+++ b/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/PanDomainAuthSettings.scala
@@ -1,20 +1,20 @@
 package com.gu.pandomainauth.model
 
-import com.gu.pandomainauth.{PrivateKey, PublicKey, PublicSettings, Secret}
+import com.gu.pandomainauth.{PrivateKey, PublicKey, Secret}
 
 case class PanDomainAuthSettings(
   secret: Secret,
   publicKey: PublicKey,
   privateKey: PrivateKey,
+  cookieSettings: CookieSettings,
   googleAuthSettings: GoogleAuthSettings,
   google2FAGroupSettings: Option[Google2FAGroupSettings]
-) {
-  @deprecated("Use com.gu.pandomainauth.PublicSettings.cookieName", "0.2.7")
-  val cookieName = PublicSettings.cookieName
+)
 
-  @deprecated("Use com.gu.pandomainauth.PublicSettings.assymCookieName", "0.2.7")
-  val assymCookieName = PublicSettings.assymCookieName
-}
+case class CookieSettings(
+  cookieName: String,
+  assymCookieName: String
+)
 
 case class GoogleAuthSettings(
   googleAuthClient: String,
@@ -31,6 +31,10 @@ case class Google2FAGroupSettings(
 object PanDomainAuthSettings{
 
   def apply(settingMap: Map[String, String]): PanDomainAuthSettings = {
+    val cookieSettings = CookieSettings(
+      cookieName = settingMap("cookieName"),
+      assymCookieName = settingMap("assymCookieName")
+    )
 
     val googleAuthSettings = GoogleAuthSettings(settingMap("googleAuthClientId"), settingMap("googleAuthSecret"))
 
@@ -43,6 +47,13 @@ object PanDomainAuthSettings{
       Google2FAGroupSettings(serviceAccountId, serviceAccountCert, adminUser, group)
     }
 
-    PanDomainAuthSettings(Secret(settingMap("secret")), PublicKey(settingMap("publicKey")), PrivateKey(settingMap("privateKey")), googleAuthSettings, google2faSettings)
+    PanDomainAuthSettings(
+      Secret(settingMap("secret")),
+      PublicKey(settingMap("publicKey")),
+      PrivateKey(settingMap("privateKey")),
+      cookieSettings,
+      googleAuthSettings,
+      google2faSettings
+    )
   }
 }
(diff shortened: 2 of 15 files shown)


pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
```diff
@@ -19,20 +19,22 @@ import scala.concurrent.duration.FiniteDuration
...
   */
 class PanDomainAuthSettingsRefresher(
   val domain: String,
   val system: String,
+  bucketName: String,
   actorSystem: ActorSystem,
-  awsCredentialsProvider: AWSCredentialsProvider = new DefaultAWSCredentialsProviderChain(),
-  awsRegion: Option[Region] = Option(Region getRegion Regions.EU_WEST_1),
+  awsCredentialsProvider: AWSCredentialsProvider,
+  awsRegion: String,
```
akash1810: why the change in type?

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/PanDomainAuthSettings.scala
```diff
@@ -31,8 +32,16 @@ case class Google2FAGroupSettings(
...
+    val cookieSettings = CookieSettings(
+      cookieName = settingMap("cookieName"),
+      assymCookieName = settingMap("assymCookieName")
+    )
 
-    val googleAuthSettings = GoogleAuthSettings(settingMap("googleAuthClientId"), settingMap("googleAuthSecret"))
+    val oAuthSettings = OAuthSettings(
+      settingMap("clientId"),
+      settingMap("clientSecret"),
+      settingMap("discoveryDocumentUrl")
```
akash1810: what would `discoveryDocumentUrl` be in a google auth context?

mbarton: `https://accounts.google.com/.well-known/openid-configuration` - see the top level PR description

akash1810: img

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
mbarton: The latest AWS SDKs deprecated the previously code we used to construct the S3 client. Instead they suggest using the `withRegion` builder which either takes a string or a Regions object which I have found fiddly to build from the Region enums themselves.

It also means we don't need an SDK bump to target new AWS regions, although I'm not sure that's important.

Not a big deal though, we can keep it as the typed enum.

pan-domain-auth-example/app/di.scala
```diff
@@ -19,15 +19,24 @@ class AppComponents(context: Context) extends BuiltInComponentsFromContext(conte
...
-  ))
+  val awsRegion = "eu-west-1"
+
+  // Change this to point to the S3 bucket containing the settings file
+  val bucketName = "pan-domain-auth-settings"
+
+  // Prefer profile credentials over static credentials
+  // If you work at the Guardian you can get these credentials from Janus
+  val awsCredentialsProvider = new AWSCredentialsProviderChain(
+    new ProfileCredentialsProvider("workflow"),
```
adamnfish: If your goal is to decouple this from Guardian tooling does this still make sense?

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PanDomain.scala
```diff
@@ -8,7 +8,7 @@ object PanDomain {
   /**
    * Check the authentication status of the provided credentials by examining the signed cookie data.
    */
-  def authStatus(cookieData: String, publicKey: PublicKey, validateUser: AuthenticatedUser => Boolean = guardianValidation): AuthenticationStatus = {
+  def authStatus(cookieData: String, publicKey: PublicKey, validateUser: AuthenticatedUser => Boolean): AuthenticationStatus = {
```
adamnfish: I think this is a pretty important change for us. If tools were relying on this default argument then their security will break when they upgrade.

It might be safer to keep this as is and introduce a new function that doesn't include our check?

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/PanDomainAuthSettings.scala
```diff
@@ -31,8 +32,16 @@ case class Google2FAGroupSettings(
 object PanDomainAuthSettings{
 
   def apply(settingMap: Map[String, String]): PanDomainAuthSettings = {
+    val cookieSettings = CookieSettings(
+      cookieName = settingMap("cookieName"),
```
adamnfish: Are both these cookies still used?

mbarton: Unfortunately yes :(

There are still a couple of tools that need upgrading

pan-domain-auth-example/app/di.scala
mbarton: I've updated the example to have a more general note of advice:

```scala
// Customise as appropriate depending on how you manage your AWS credentials
```

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PanDomain.scala
mbarton: Hmm I'm torn. You're right, the safest option is not to change behaviour but the existing code is confusing because there are essentially two `validateUser` functions: one you can customise and this one that you can't. 

I found this code in [mobile-entry](https://github.com/guardian/mobile-entry/blob/master/app/ext/EntryAuthActions.scala#L112) that works-around this in a pretty ugly way.

Adding a third function would increase the tech debt in the library, although I guess we could minimise it by adding deprecated annotations.

Can we catch up quickly IRL about this? I've done an audit of our `validateUser` functions on GitHub so that can feed into the discussion.

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/PanDomainAuthSettings.scala
adamnfish: *facepalm

It might be nice to do that upgrade ahead of this so we can have v2 be just the new cookie (ideally with the spelling mistake fixed). Maybe that's too optimistic.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PanDomain.scala
adamnfish: Yep, let's converse in the real world.

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
sihil: The SDK advantage doesn't exist. When you pass in a string the SDK will find the Regions object for you so will hit the same problem. 

If you wanted to use a region not in the SDK you need to use withEndpoint instead of withRegion and pass in the endpoint URL.

I favour types here.

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/PanDomainAuthSettings.scala
sihil: Am I right in thinking that we only drop the assym cookie? Might be worth renaming these concepts to `legacyCookie`?

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
rtyley: :+1: for types! I think though that you might actually want to change the type like this:

```suggestion
  awsRegion: Option[com.amazonaws.regions.Regions],
```

...note that the type is now `Regions` rather than `Region` (and you get `Regions.EU_WEST_1` etc pre-constructed for you!) - and this object is accepted by the newer `com.amazonaws.client.builder.AwsClientBuilder#withRegion(com.amazonaws.regions.Regions)` builder method.

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/PanDomainAuthSettings.scala
mbarton: We do still [drop the legacy cookie](https://github.com/guardian/pan-domain-authentication/blob/master/pan-domain-auth-play_2-6/src/main/scala/com/gu/pandomainauth/action/Actions.scala#L167). I have renamed the setting as suggested.

I will merge this PR, upgrade the last two remaining tools that still rely on the old cookie and then raise a follow up PR to remove the code just leaving the asymmetric one.

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
mbarton: Thanks everyone, makes sense. I've followed Roberto's suggestion but made the region mandatory. Any default region should be passed in from the calling application code.

mbarton:
Thanks for the reviews everyone. I am going to merge this now but will not do a release immediately.

After talking to Adam and Simon we agreed that it would be prudent to do some additional work before others use this library in anger. The steps I plan to take before doing a release are:

- Update the remaining two applications that rely on the legacy non-asymmetric cookie
- Remove the legacy cookie entirely from the library
- Verify the crypto calls and algorithms we are using are still those recommended (including running it past infosec)

Merging now gives us the option to do a release without those steps should it be required.

PR #46, opened 2018-12-27
Update bouncy castle to latest version

mchv:
@guardian/digital-cms

Current version is vulnerable to Insecure Encryption. It has a flaw in the Low-level interface to RSA key pair generator, specifically RSA Key Pairs generated in low-level API with added certainty may have less M-R tests than expected.

[CVE](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-1000180)

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -26,7 +26,7 @@ object Dependencies {
   )
 
   val cryptoDependencies = Seq(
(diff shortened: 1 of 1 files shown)


mchv:
I have just published `0.7.2` version with those changes.
I may create one or two PR as example on how to update.

PR #47, opened 2019-01-14
Remove legacy cookie

mbarton:
Remove the legacy cookie that required knowledge of the private key for verification. Moving forward we now only support the asymmetric cookie that can be verified with the public key.


pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/PanDomainAuthSettings.scala
```diff
@@ -33,8 +32,8 @@ object PanDomainAuthSettings{
 
   def apply(settingMap: Map[String, String]): PanDomainAuthSettings = {
     val cookieSettings = CookieSettings(
-      legacyCookieName = settingMap("legacyCookieName"),
-      assymCookieName = settingMap("assymCookieName")
+      // Try legacy setting name first for compatibility
+      cookieName = settingMap.getOrElse("assymCookieName", settingMap("cookieName"))
```
franziskas: I think this could be clearer if `assymCookieName` was a constant named `legacySettingName` or similar

PR #48, opened 2019-01-14
Pre 1.0 tidy up

### LINKED ISSUE #19: Static name of S3 bucket: pan-domain-auth-settings

mbarton:
Now Panda is not dependent on a particular OAuth provider (#45), simplifying the setup of [the Grid](https://github.com/guardian/grid) for non-Guardian users, I thought it was a good time to go through tidying up the API since we were going to release a breaking change anyway.

The journey also led me to remove some unnecessary code and ibrary dependencies and refactor some settings parsing.

Apologies in advance, this is a proper grab-bag of a PR. I tried splitting it up into lots of small changes but found it challenging due to some of them being dependent on each other. Instead I've gone through the PR adding comments describing my reasons for the changes.

I've also made some sequencing and wording changes to the README. They are perhaps best reviewed by just reading through the new version to see if it makes sense rather than looking at the diff.

Fixes #19


README.md
```diff
@@ -35,144 +48,69 @@ Pan domain auth is split into 4 modules.
...
+The `pan-domain-auth-core` library provides the core utilities to load settings, create and validate the cookie and
+check if the user has mutli-factor auth turned on when usng Google as the provider.
 
 The `pan-domain-auth-play_2-6` library provide an implementation for play apps. There is an auth action that can be applied to the
 endpoints in your application that will do checking and setting of the cookie and will give you the OAuth authentication mechanism
 and callback. This is the only framework specific implementation currently (due to play being the framework predominantly used at The
 Guardian), this can be used as reference if you need to implement another framework implementation. This library is for applications
 that need to be able to issue and verify logins which is likely to include user-facing applications.
 
-The `pan-domain-auth-example` provides an example app with authentication. This is implemented in play and is used for testing.
```
mbarton: The main change here is to move the `Requirements` section about the `Integrating with your app` section, with a link to skip if you already have everything configured. Users setting up the Grid for the first time found this confusing.

I have also moved code examples to the examples project wherever possible so they are more likely to stay up to date with future changes (because breaking changes would cause compile errors etc).

The README is still a bit spaghetti but I have hopefully removed anything that is no longer true and tidied it up to be a bit more comprehensible

README.md
```diff
@@ -351,82 +210,8 @@ class Application(
...
-your framework and http client etc.
-
-More examples and framework clients may be added in the future as they become available.
-
-### Configuring access to the S3 bucket
-
-Access to the S3 bucket is controlled by passing an [AWSCredentialsProvider](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/AWSCredentialsProvider.html)
-and [region](https://docs.aws.amazon.com/general/latest/gr/rande.html) to the `PanDomainAuthSettingsRefresher`.
-
-### The user object
```
mbarton: I didn't see much value in describing the fields of `User` and `AuthenticatedUser`. They are self-descriptive and covered in the example code.

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
```diff
@@ -1,83 +1,61 @@
...
   val domain: String,
   val system: String,
-  bucketName: String,
-  actorSystem: ActorSystem,
-  awsCredentialsProvider: AWSCredentialsProvider,
-  awsRegion: Regions,
-  proxyConfiguration: Option[ProxyConfiguration] = None
+  val bucketName: String,
+  settingsFileKey: String,
+  val s3Client: AmazonS3,
```
mbarton: This change to simply take an existing S3 client means we have much more scope for integrating with other environments. It also means we no longer need the credentials provider, region or proxy configuration since they are set on the builder used to create the client (ie in calling code)

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
```diff
@@ -1,83 +1,61 @@
...
+  scheduler: ScheduledExecutorService = Executors.newScheduledThreadPool(1)
```
mbarton: By using a native JDK class here to update the settings we remove a dependency on Akka. Users can still pass in their own scheduler should they wish (to customise thread names for example)

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
```diff
@@ -1,83 +1,61 @@
...
 ) {
-  lazy val bucket = new S3Bucket(bucketName, awsCredentialsProvider, awsRegion, proxyConfiguration)
-
-  private lazy val settingsMap = bucket.readDomainSettings(domain)
-  private lazy val authSettings: Agent[PanDomainAuthSettings] = Agent(PanDomainAuthSettings(settingsMap))
```
mbarton: Akka agents are deprecated so I have replaced this with a straight-forward AtomicReference and removed the dependency

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
```diff
@@ -1,83 +1,61 @@
...
+  private val authSettings: AtomicReference[PanDomainAuthSettings] = new AtomicReference[PanDomainAuthSettings](loadSettings() match {
+    case Right(settings) => PanDomainAuthSettings(settings)
+    case Left(err) => throw Settings.errorToThrowable(err)
+  })
 
-  override def receive: Receive = {
-    case Refresh => {
-      try {
-        val settingsMap = bucket.readDomainSettings(domain)
+  scheduler.scheduleAtFixedRate(() => refresh(), 1, 1, TimeUnit.MINUTES)
```
mbarton: This refresh time was not customisable by calling code before, but it is for the public settings refresher! Perhaps we should also fix in this release?

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
```diff
@@ -1,83 +1,61 @@
...
+    Settings.fetchSettings(settingsFileKey, bucketName, s3Client).flatMap(Settings.extractSettings)
   }
 
-  override def postRestart(reason: Throwable) {
-    reschedule
-  }
+  private def refresh(): Unit = {
+    loadSettings() match {
+      case Right(settings) =>
+        logger.info(s"Updated pan-domain settings for $domain")
```
mbarton: Wording changed on these logging messages for consistency with the public settings refresher

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/service/Google2FAGroupChecker.scala
```diff
@@ -10,7 +11,7 @@ import com.google.api.services.admin.directory.{Directory, DirectoryScopes}
 import scala.collection.JavaConverters._
 import com.gu.pandomainauth.model.{AuthenticatedUser, Google2FAGroupSettings}
 
-class GroupChecker(config: Google2FAGroupSettings, bucket: S3Bucket) {
```
mbarton: I deleted the S3Bucket class since it provided no value other than very minimal wrappers around the S3 client

pan-domain-auth-example/app/VerifyExample.scala
```diff
@@ -0,0 +1,43 @@
+import com.amazonaws.auth.DefaultAWSCredentialsProviderChain
+import com.amazonaws.regions.Regions
+import com.amazonaws.services.s3.AmazonS3ClientBuilder
+import com.gu.pandomainauth.model.{Authenticated, AuthenticatedUser, GracePeriod}
+import com.gu.pandomainauth.{PanDomain, PublicSettings}
+
+object VerifyExample {
```
mbarton: This replaces the example code from the README

pan-domain-auth-example/app/controllers/ExampleAuthActions.scala
```diff
@@ -14,6 +14,9 @@ trait ExampleAuthActions extends AuthActions {
     PanDomain.guardianValidation(authedUser)
   }
 
+  // By default, the user validation method is called every request. If your validation
```
mbarton: This was moved from the README

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala
```diff
@@ -1,83 +1,61 @@
...
 class PanDomainAuthSettingsRefresher(
   val domain: String,
   val system: String,
-  bucketName: String,
-  actorSystem: ActorSystem,
-  awsCredentialsProvider: AWSCredentialsProvider,
-  awsRegion: Regions,
-  proxyConfiguration: Option[ProxyConfiguration] = None
+  val bucketName: String,
+  settingsFileKey: String,
```
mbarton: Previously the settings file was determined by naming convention from the domain eg `<domain>.settings`. I think it's easier for new users of the library if they pass the name of the file in the bucket directly. The README still encourages them to use the naming convention.

I guess this does require the same key to be copy-pasted across all apps using Panda so perhaps the naming convention is better?

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
```diff
@@ -1,88 +1,56 @@
...
- * @param ec         Implicit execution context used to fetch the settings
+ * Class that contains the static public settings and includes mechanism for fetching the public key. Once you have an
+ * instance, call the `start()` method to load the public data.
+  *
+  * @param settingsFileKey the settings file for the domain in the S3 bucket (eg local.dev.gutools.co.uk.public.settings)
+  * @param bucketName      the name of the S3 bucket (eg pan-domain-auth-settings)
+  * @param s3Client        the AWS S3 client that will be used to download the settings from the bucket
+  * @param scheduler       optional scheduler that will be used to run the code that updates the bucket
  */
-class PublicSettings(domain: String, bucketName: String, callback: Try[Map[String, String]] => Unit = _ => (),
```
mbarton: `domain` was removed since it was unused.

`callback` has also been removed since it is only used in one place across all the code in the Guardian: [s3-uploader for logging](https://github.com/guardian/s3-upload/blob/master/app/lib/PanAuth.scala#L19). The logging is useful for everyone so I've added it to the library itself.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
```diff
@@ -1,88 +1,56 @@
...
-                     scheduler: Scheduler = StdSchedulerFactory.getDefaultScheduler())
```
mbarton: As with the other settings refresher, using this JDK scheduler allows us to remove a dependency (in this case Quartz)

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
```diff
@@ -1,88 +1,56 @@
...
-                    (implicit client: OkHttpClient, ec: ExecutionContext) {
+class PublicSettings(settingsFileKey: String, bucketName: String, s3Client: AmazonS3,
```
mbarton: `PublicSettings` previously made an HTTP request to retrieve the settings file. This added *another* dependency (OkHttp) and as far as I could see provided no value over using the normal S3 client other than the request being asynchronous. The only piece of code that [reads the public settings directly](https://github.com/guardian/content-api/blob/99cf94c31ae3b3dba4618b30cce6ac90c636286a/gateways/panda-lambda/src/main/scala/com/gu/pandalambda/Lambda.scala#L30) added an `Await` around it anyway!

I think it's clearer and more consistent if we use an S3 client in this class.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
```diff
@@ -91,56 +59,17 @@ object PublicSettings {
...
 
-    client.newCall(req).enqueue(new Callback {
-      override def onFailure(call: Call, e: IOException): Unit = promise.success(Left(e))
-      override def onResponse(call: Call, response: Response): Unit = promise.success(Right(response.body.string))
-    })
-
-    promise.future
-  }
-
-  private[pandomainauth] def extractSettings(settingsAttempt: Either[Throwable, String]): Future[Map[String, String]] =
```
mbarton: These methods have been moved to the `Settings` object so they can be shared with the other refresher

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/Settings.scala
```diff
@@ -0,0 +1,66 @@
...
+
+sealed trait SettingsFailure
+case class SettingsDownloadFailure(cause: Throwable) extends SettingsFailure
+case class SettingsParseFailure(cause: Throwable) extends SettingsFailure
+case object PublicKeyFormatFailure extends SettingsFailure
+case object PublicKeyNotFoundFailure extends SettingsFailure
+
+object Settings {
+  // internal functions for fetching and parsing the responses
+  def fetchSettings(settingsFileKey: String, bucketName: String, s3Client: AmazonS3): Either[SettingsFailure, String] = try {
```
mbarton: Both the normal settings refresher and the public settings refresher parsed their settings in very similar ways so I have refactored them to use the same code here.

I have also changed to using an `Either` type rather than exceptions. The existing behaviour for each settings refresher is maintained though. The public settings refresher never throws an exception whereas the normal refresher throws an exception (created by wrapping the Failure) if the settings cannot be loaded at construction.

project/Dependencies.scala
```diff
@@ -47,11 +41,5 @@ object Dependencies {
...
   }
 
-  val logback: Seq[ModuleID] = {
-    val version = "1.2.3"
-    Seq(
-      "ch.qos.logback" % "logback-core" % version,
-      "ch.qos.logback" % "logback-classic" % version
-    )
-  }
+  val loggingDependencies = Seq("org.slf4j" % "slf4j-api" % "1.7.25")
```
mbarton: The general practice for libraries on the JVM is to depend on a logging API rather than an implementation, hence the chanage here. Again this will reduce the possibility of dependency hell for applications consuming the library.

PR #50, opened 2019-01-14
Move main build definition to build.sbt

mbarton:
This **hopefully** means Snyk can start to understand it and generate reports

@guardian/digital-cms


build.sbt
```diff
@@ -0,0 +1,147 @@
...
+import play.sbt.routes.RoutesKeys._
+
+val scala211 = "2.11.12"
+val scala212 = "2.12.4"
+
+val commonSettings =
+  Seq(
+    scalaVersion := scala212,
+    scalaVersion in ThisBuild := scala212,
+    crossScalaVersions := Seq(scala211, scala212),
```
akash1810: I suspect we can remove 2.11 now, if not very soon.

mbarton: I have a spreadsheet (internal to the Guardian) that lists all the versions of Panda that are still out there in the wild. I'll share it with you now

PR #54, opened 2019-03-11
Port across node implementation

mbarton:
Port across the implementation from https://github.com/guardian/pan-domain-node, upgrading it to Typescript in the process. This is so I can consume it from a typescript project (https://github.com/guardian/editorial-tools-platform/pull/248). As the code was straight forward I felt a full port made more sense than simply adding a typescript definition file.

I have merged in all the functionality from the micro-libraries https://github.com/guardian/pan-domain-validate-user and https://github.com/guardian/pan-domain-public-keys, with the intention of eventually replacing their use and removing them.

I will also raise a follow up PR to port across https://github.com/guardian/pan-domain-lambda as well.

TODO
- [ ] add to CI build


pan-domain-node/src/api.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export interface AuthenticationResult {
+    status: AuthenticationStatus,
+    user?: User 
+}
+
+export type ValidateUserFn = (user: User) => boolean;
+
+export function guardianValidation(user: User): boolean {
```
akash1810: would `isGuardianUser` be more intention revealing?

pan-domain-node/src/panda.ts
```diff
@@ -0,0 +1,111 @@
...
+import { httpGet, base64ToPEM, parseCookie, verifySignature, parseUser } from './utils';
+import { ValidateUserFn, AuthenticationStatus, User, AuthenticationResult } from './api';
+
+interface PublicKeyHolder {
+    key: string,
+    lastUpdated: Date
+}
+
+function fetchPublicKey(region: string, bucket: String, keyFile: String): Promise<PublicKeyHolder> {
+    const path = `https://s3.${region}.amazonaws.com/${bucket}/${keyFile}`;
```
akash1810: Does this mean this file is publicly accessible? 😱

pan-domain-node/src/panda.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    getPublicKey(): Promise<string> {
+        return this.publicKey.then(({ key, lastUpdated }) => {
+            const now = new Date();
+            const diff = now.getMilliseconds() - lastUpdated.getMilliseconds();
+
+            if(diff > this.keyCacheTime) {
+                this.publicKey = fetchPublicKey(this.region, this.bucket, this.keyFile);
```
akash1810: Why are we re-assigning here? We're already setting this property in the constructor.

pan-domain-node/src/utils.ts
```diff
@@ -0,0 +1,122 @@
...
+export function base64ToPEM (key: string): string {
+	const tmp = [];
+
+    const ret = [new Buffer(PEM_HEADER).toString('ascii')];
+
+    for (let i = 0, len = key.length; i < len; i++) {
+
+        if (i>0 && i%64 === 0) {
+            ret.push(tmp.join(''));
+            tmp.length = 0;
```
akash1810: 🤔 nice

pan-domain-node/src/utils.ts
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        tmp.push(key[i]);
+
+        if (i === key.length - 1) {
+            ret.push(tmp.join(''));
```
akash1810: minor: we could do this outside the loop, removing the `if`

pan-domain-node/src/utils.ts
```diff
@@ -0,0 +1,122 @@
...
+                    
+                    reject(error);
+                }
+            });
+        });
+    });
+}
+
+export function parseUser(data: string): User {
+    const params = qs.parse(data);
```
akash1810: could we use `URLSearchParams` here to drop the dependency on `qs`?

pan-domain-node/src/utils.ts
```diff
@@ -0,0 +1,122 @@
...
+        if(!value) { throw new Error(`Missing ${name}`) }
+
+        return value.split(",");
+    }
+
+    return {
+        firstName: stringField("firstName"),
+        lastName: stringField("lastName"),
+        email: stringField("email"),
+        avatarUrl: <string>params["avatarUrl"],
```
akash1810: why not use `stringField` here?

pan-domain-node/test/fixtures.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const encodedPublicKey = "MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEAur0hOjhB2QWjwOopCR+Qo27AYv97BJkVaKWPXpj9RfvY1wtpIratDN6tkXN9WCRPzVX8+5qaW034Kvf9WwBZD1ntS8iHYwY1YUaU8Mrp2sRT3K0RqyBlTswIH3HIqpASqv7ZtwDHdhk7Cbd13P5aomJSOjYFhCDUi3sRbjJP1kb6uQLdkZj8fIU518HzSR7Kw7p2mbDqSrGbnaeHWd0Tr3BvDHp9Pi0KpSAVm2qWAXix+BcjMA4ar7kLU1Pre0lt4K4DlSvq5XoHdX9/yvS6KGf+8pXDR9bY6dgRPSG4mzKpiKfkv1eXE8WKs0q3217QZItaSjocw4d0o47vSN+/MAh9V5Zewyb6ogs8JicX3Y3FPG29I1g8iLf3kBZ7V4mUimGuOq/L+1YVvyOTb2zWWMjNmECO2lrxXJc5LWRs5FmSJyCdilRktDE0WTGUo89O+DcF2752qtpUmlV2fllU1LXIAn0TJKiAZspKakamifrgYFIzZK4oZ8wDeFesQB/a/U7wtyv85vknzCtMLI28dnpGQ/ZFHNqWYVaHoHsnEmWion7lgMctnpY5pwKFfUSfZecl2Xqwjk1HZj71A9TFNQj+/x4z957cNtx+utAkGinK3eZF+H1o5YnSgjg4hN41kbXttk8nADerPdF7hDS6np7xzUl6qOicJhEOJ5x0c18CAwEAAQ==";
+export const publicKey = base64ToPEM(encodedPublicKey);
+
+/*
```
akash1810: why are these commented out? should we remove them?

pan-domain-node/src/panda.ts
```diff
@@ -0,0 +1,111 @@
...
+                key: base64ToPEM(config.publicKey),
+                lastUpdated: new Date()
+            };
+        } else {
+            throw new Error("Missing publicKey setting from config");
+        }
+    });
+}
+
+export function verifyUser(pandaCookie: string | undefined, publicKey: string, now: number, validateUser: ValidateUserFn): AuthenticationResult {
```
akash1810: `now` isn't an especially clearly named variable, that is, `now` doesn't lend it self to be assignable

pan-domain-node/test/panda.test.ts
```diff
@@ -0,0 +1,28 @@
...
+
+test("return invalid cookie for a malformed signature", () => {
+    const [data, signature] = sampleCookie.split(".");
+    const testCookie = data + ".1234";
+
+    expect(verifyUser(testCookie, publicKey, 0, guardianValidation).status).toBe(AuthenticationStatus.INVALID_COOKIE);
+});
+
+test("return expired", () => {
+    expect(verifyUser(sampleCookie, publicKey, 5678, guardianValidation).status).toBe(AuthenticationStatus.EXPIRED);
```
akash1810: could we remove the magic number?

pan-domain-node/tsconfig.json
```diff
@@ -0,0 +1,60 @@
+{
```
akash1810: do we need all the comments in this file?

pan-domain-node/src/api.ts
mbarton: Possibly. I named it the same as the [function in the Scala library](https://github.com/guardian/pan-domain-authentication/blob/master/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PanDomain.scala#L52) just for API consistency

pan-domain-node/src/panda.ts
mbarton: Yep and that's fine because it's assymetric crypto, like how PGP works.

Anyone can verify the signature using the public key but only those who have the private key can generate the signature.

pan-domain-node/src/panda.ts
mbarton: If the public key is older than the cache time we should re-download it in case it has changed

pan-domain-node/src/utils.ts
mbarton: Copy paste from the [format-pem-keys](https://github.com/guardian/format-pem-keys) microlibrary and I was fearful to touch the code!

pan-domain-node/src/utils.ts
mbarton: Definitely! I've done that

pan-domain-node/src/utils.ts
mbarton: It's an optional field. The code is clearer now we've moved to `URLSearchParams`

pan-domain-node/test/fixtures.ts
mbarton: They're the Scala case class representation of the data contained in the encoded cookie below. I left them in to avoid confusion

pan-domain-node/test/panda.test.ts
mbarton: No. It's a test for expiry where the passed in timestamp (the time "now" when called in real code outside of test)

pan-domain-node/src/panda.ts
mbarton: I've renamed it to `currentTimestamp`

pan-domain-node/tsconfig.json
mbarton: It's the default you get with `tsc --init`. I'm not a big fan myself so I've removed them

pan-domain-node/src/utils.ts
mbarton: I've change `tmp` to be a `var` instead

mbarton:
@akash1810 would you be able to have another look after my review fixes?

pan-domain-node/.nvmrc
```diff
@@ -0,0 +1 @@
+6
```
akash1810: worth targeting a higher version of node?

pan-domain-node/src/api.ts
akash1810: Don't like it, but the reasoning makes sense.

pan-domain-node/src/utils.ts
```diff
@@ -0,0 +1,120 @@
...
+        });
+    });
+}
+
+export function parseUser(data: string): User {
+    const params = new URLSearchParams(data);
+
+    function stringField(name: string): string {
+        const value = params.get(name);
+        if(!value ) { throw new Error(`Missing  ${name}`) }
```
akash1810: minor formatting issue here

pan-domain-node/test/fixtures.ts
akash1810: could we add a comment to that effect?

pan-domain-node/test/panda.test.ts
akash1810: Would it be worth using a `const` here to improve readability? `const someTimeInTheFuture = 5678`

pan-domain-node/.nvmrc
mbarton: Yep definitely now Node 6 lambdas are EOL

PR #58, opened 2019-05-02
Play 2.7

mbarton:
No major changes, just a copy-paste job from 2.6 and editing the log messages to use an SLF4j logger instance since the main Play `Logger` object is deprecated in 2.7.

I've also removed some transitive dependencies that seemed to be unused (it still compiles and the tests pass). Finally I've removed some old packaging stuff for the example app which was not required.


akash1810:
can we share the code rather than copy+paste?

mbarton:
Yeah I've thought about that before and I don't think so - since there's minor changes that need to be made for compatibility with each version. Java doesn't have a pre-processor so we couldn't use the traditional C/C++ approach of `#ifdef` for this.

Maintaining a set of patch files applied by scripts is possible but feels more fiddly the copy-paste.

Very open to suggestions though

PR #61, opened 2019-08-09
made email and email verified from jwt response optional, if not present get email from user info response

jamie-pitts:
At the moment the oauth callback is expecting an `email` and `email_verified` field in the JWT payload from the oauth provider, however these fields arent part of the standard spec ( https://en.wikipedia.org/wiki/JSON_Web_Token#Standard_fields ). 

This changes it so that it will try and get it from the jwt token and then if its not present get it from the userInfo response.

I have tested with both google and our own oauth provider (BBC Login)

### DIFF
--- a/pan-domain-auth-play_2-6/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
+++ b/pan-domain-auth-play_2-6/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
@@ -85,7 +85,7 @@ class OAuth(config: OAuthSettings, system: String, redirectUrl: String) {
                   user = User(
                     userInfo.given_name,
                     userInfo.family_name,
-                    jwt.claims.email,
+                    jwt.claims.email.getOrElse(userInfo.email),
                     userInfo.picture
                   ),
                   authenticatingSystem = system,
--- a/pan-domain-auth-play_2-6/src/main/scala/com/gu/pandomainauth/service/oAuthModel.scala
+++ b/pan-domain-auth-play_2-6/src/main/scala/com/gu/pandomainauth/service/oAuthModel.scala
@@ -30,8 +30,8 @@ object Token {
   }
 }
 
-case class JwtClaims(iss: String, sub: String, azp: Option[String], email: String, at_hash: String, email_verified: Boolean,
-                     aud: String, hd: Option[String], iat: Long, exp: Long)
+case class JwtClaims(iss: String, sub: String, azp: Option[String], email: Option[String], at_hash: String,
+                     email_verified: Option[Boolean], aud: String, hd: Option[String], iat: Long, exp: Long)
 object JwtClaims {
   implicit val claimsReads = Json.reads[JwtClaims]
 }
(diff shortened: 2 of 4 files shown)


akash1810:
Looks good, however I've not got enough knowledge about this library to be able to confidently approve.

sihil:
If `userInfo.email` is part of the spec and exists in google then I'd suggest we switch to consume only `userInfo` rather than a mix of that and the `jwtClaim`. Only issue would be if that e-mail is not the canonical e-mail (at the Guardian we have multiple domains but many of our systems assume a particular domain).

Also worth mentioning that unique user identification is better done with `sub` (which we've now switched to in our permissions app) rather than e-mail.

mbarton:
[The User Info endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) is standard but what claims is returns is not, so there could be some providers out there that don't return an email.

We do request the `email` scope however so I think it's fine for us to rely on it being there:

https://github.com/guardian/pan-domain-authentication/blob/c6b7d3dbfb2d287813bb0c4b9b0f23568533feb7/pan-domain-auth-play_2-6/src/main/scala/com/gu/pandomainauth/service/OAuth.scala#L52

I think the requested scopes should be returned in the token, not just the user info response? The standard [seems to hint at it](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) but I guess your provider does not @jamie-pitts?

> Also worth mentioning that unique user identification is better done with sub (which we've now switched to in our permissions app) rather than e-mail.

I guess we could switch to it even though the resulting field in the library is called `email`? Ultimately Panda is just to make login work for tools so I'm not too fussed providing it's true in reality.

pan-domain-auth-play_2-6/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
```diff
@@ -85,7 +85,7 @@ class OAuth(config: OAuthSettings, system: String, redirectUrl: String) {
                   user = User(
                     userInfo.given_name,
                     userInfo.family_name,
-                    jwt.claims.email,
+                    jwt.claims.email.getOrElse(userInfo.email),
```
mbarton: This needs copying into the `pan-domain-auth-play_2-7` as well. Maintaining two copies is rubbish so any suggestions welcome for how to avoid duplication.

jamie-pitts:
Thanks for the comments. @akash1810 @sihil @mbarton I've updated the play 2.7 library now. 

Agreed it's a bit confusing if the email should be included in the jwt payload if you request the scope for email, it looks like its up to the provider if they want to include it in either the payload or the userinfo response. This current implementation will guarantee it should be able to receive back an email at least, I would rather keep it this way for now because I wouldn't want it to have any negative effects on your system in the case the userinfo endpoint may not return the same email as the jwt payload for whatever reason. 

In terms of changing to use `sub`, that probably makes sense but should maybe be added as an extra field and re-worked as a separate issue, as currently I believe if you were to change the email to sub, then the grid probably wouldn't show users uploads when they click on 'my uploads'?

sihil:
You want to review as well @mbarton? Happy to merge and release once you have.

sihil:
Released as part of `0.9.1`.

PR #73, opened 2020-05-21
Add Play 2.8 artifact

mbarton:
The code is a copy paste from Play 2.7 which means that version may well work directly with Play 2.8 but I think it's clearer if we publish a separate artifact anyway.

Further updates I think are worth doing in separate PRs:
- Cross build against Scala 2.12 and 2.13
- Upgrade to SBT v1
- General library version bumps

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -20,6 +20,14 @@ object Dependencies {
     )
   }
 
+  val playLibs_2_8 = {
+    val version = "2.8.1"
+    Seq(
+      "com.typesafe.play" %% "play" % version % "provided",
+      "com.typesafe.play" %% "play-ws" % version % "provided"
+    )
+  }
+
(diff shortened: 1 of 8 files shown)


project/Dependencies.scala
```diff
@@ -20,6 +20,14 @@ object Dependencies {
     )
   }
 
+  val playLibs_2_8 = {
+    val version = "2.8.1"
+    Seq(
+      "com.typesafe.play" %% "play" % version % "provided",
```
akash1810, line 26: ooh, what does `provided` do?

mbarton: It indicates to the package manager that projects using this library will provide their own version of the package rather than have it download it as a transitive dependency of this library.

I think we use it here to avoid problems when Play is brought in via the SBT plugin in projects that use Panda

PR #77, opened 2021-02-16
setApplicationName in credential builder to stop it complaining

AWare:
see #60 

@guardian/digital-cms


AWare:
this doesn't build, will revisit when we update library in grid

PR #115, opened 2023-03-15
Bump sbt to 1.7.2

jonathonherbert:
## What's changed?

Bumps sbt to 1.7.2 – >1.4.x is necessary for the project to compile on M1 machines. I've also bumped the relevant Play libs, as Play 2.8 has problems with later versions of sbt that are [fixed in >2.8.8.](https://github.com/sbt/sbt/issues/6400)

## Why not latest (1.8.2 at time of writing)?

Bumping to latest pulls in two different major versions of scala-xml. There are [workarounds](https://github.com/sbt/sbt/issues/7007#issuecomment-1221823661) that IIUC involve handwaving this conflict, but given that this is an important dependency that touches many projects, perhaps that's best avoided.

[The latest version of Twirl uses scala-xml 2.x](https://github.com/playframework/twirl/blob/df5f7ef869f78b3da7109edee4966a8ec16bb19d/build.sbt#L89), which I think will be coming in Play 2.9, so I think we'll have to wait for Play 2.9 to resolve this without that workaround. If others would prefer us to bump sbt rather than wait, we can use the workaround above.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,10 +12,10 @@ val scala213 = "2.13.8"
 val commonSettings =
   Seq(
     scalaVersion := scala212,
-    scalaVersion in ThisBuild := scala212,
+    ThisBuild / scalaVersion  := scala212,
     crossScalaVersions := Seq(scalaVersion.value, scala213),
     organization := "com.gu",
-    fork in Test := false,
+    Test / fork := false,
     resolvers ++= Seq("Typesafe Repository" at "https://repo.typesafe.com/typesafe/releases/"),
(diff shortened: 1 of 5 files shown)


andrew-nowak (approved):

rebecca-thompson (commented):
👋🏽  Since https://github.com/guardian/pan-domain-authentication/pull/119, this repository is being built using GitHub Actions. Please rebase this branch against main to ensure it's built.

Or close the PR if it's no longer needed.

andrew-nowak:
superceded in 91836c9649b4bb8b41e4a5bd061b99734ba609a0

PR #116, opened 2023-05-24
Resolve high jackson-databind vulnerabilities

rebecca-thompson:
@guardian/digital-cms

This resolves some high vulnerabilities in the Jackson libraries which are brought in transitively via the AWS SDK. 

At least some of the vulnerabilities have been resolved on the AWS-side (https://github.com/aws/aws-sdk-java/issues/2726), so we can stop overriding these libraries and bump the AWS dependency instead.

Running `snyk test` locally shows the high vulnerabilities relating to jackson disappear when bumping the AWS SDK version and removing the manual overrides.

## How to test

Publish the library locally with `sbt "+publishLocal"`, and bump to the snapshot version in a consuming project. Does authentication work as expected?

This has been tested by publishing the library locally and logging into local Composer.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -65,7 +65,6 @@ lazy val panDomainAuthVerification = project("pan-domain-auth-verification")
       ++= cryptoDependencies
       ++ awsDependencies
       ++ testDependencies
-      ++ jackson
       ++ loggingDependencies
       ++ scalaCollectionCompatDependencies,
     publishArtifact := true
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,7 +2,7 @@ import sbt._
 
 object Dependencies {
 
-  val awsDependencies = Seq("com.amazonaws" % "aws-java-sdk-s3" % "1.11.835")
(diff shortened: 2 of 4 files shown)


davidfurey:
What version of Jackson do we use now? Is the line "Akka depends on the Jackson Scala integration and requires < 2.11" still a concern?

rebecca-thompson:
> What version of Jackson do we use now? Is the line "Akka depends on the Jackson Scala integration and requires < 2.11" still a concern?

This puts us on 2.12.7 which evicts versions 2.10.* brought in by Play 2.8. I've run it against several projects and it still works with the evicted version. It it a big jump in versions though which could definitely lead to unexpected behaviour. This could be resolved by also bumping the play version to 2.8.19 which brings the Jackson dependency used by Akka/Play up to 2.12.6.

davidfurey:
If it is easy to upgrade Play to 2.8.19 I think that is worth doing to avoid potentially undiscovered runtime errors

davidfurey (approved):

PR #117, opened 2023-05-26
allow setting a cookie to force expiry early for debugging purposes

andrew-nowak:
@guardian/digital-cms

## What does this change

This sets panda to watch out for a new cookie, `panda-forceExpiry`, which if set to a non-zero value, will make the panda library treat the presented panda cookie as if it had expired

## Why

There's been a few problems around quietly re-authenticating expired users onto tools recently, and I expect more to come in the future. Each time it has been frustrating to attempt to reproduce, because cookies only expire after 1hr. We can try to delete the cookie from our browser, but this presents the same as an entirely new user, which is a different and less problematic scenario. (An entirely new user should arrive onto a page via a page transition and receive a cookie then, whereas reauthenticating a user will often need to be done in response to an XHR, which browsers apply slightly different cookie security rules.

## Does this reduce security?

This is the important question. I believe it doesn't, as the only power this new cookie has to offer is to expire a valid cookie earlier, and only if the tool in question has been upgraded to this version of panda. This cookie also is not set by the panda library, so in effect requires to be set by a developer from devtools when debugging how an app will respond to an expired cookie.

- [x] Tested locally

### DIFF
--- a/pan-domain-auth-example/app/VerifyExample.scala
+++ b/pan-domain-auth-example/app/VerifyExample.scala
@@ -44,7 +44,7 @@ object VerifyExample {
   val cacheValidation = false
 
   // To verify, call the authStatus method with the encoded cookie data
-  val status = PanDomain.authStatus("<<cookie data>>>", publicKey, validateUser, apiGracePeriod, system, cacheValidation)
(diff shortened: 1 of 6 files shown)


jonathonherbert (approved):
Looks fairly straightforward, great idea! Nice work.

I haven't had time to test this with a consuming application – unsure if you have?

andrew-nowak:
@jonathonherbert thanks! Yes I did test this in an app locally

PR #118, opened 2023-05-29
Bad request for missing cookie

davidfurey:
DevX recommended modifying this library to serve a BadRequest (400) in the case of a missing cookie, instead of throwing an exception. This will avoid services looking like they are failing their availability SLI when users are making bad requests.


jacobwinch (approved):
Looks good to me 👍

andrew-nowak (approved):
this is about the anti-forgery and login-origin cookies (trying to avoid the `.getOrElse(throw new OAuthException("missing <>"))`, not the panda cookie itself right? not opposed to this change, but I'm surprised that there's a noticeable amount of errors from these. Unless some automated system is repeatedly trying to hit our oauth callback endpoints, these cookies should always be set by the flow; if they're not then we'd probably need to look closer at why these are throwing (because any one of these exceptions means that users aren't getting reauthed)

davidfurey:
It seems there was an automated system making this request to tag manager. I'll send you a doc with more info

PR #121, opened 2023-10-31
Build for play 2.9 and 3.0; drop play 2.7

andrew-nowak:
panda will be one of the first blockers to trying a play upgrade on another tool

~I haven't yet tried a snapshot of this, which will be next step before requesting review and publishing this~ released as 1.4.0-SNAPSHOT

Also includes a bit of a restructure of the repo - updated build.sbt, trying a hack to remove the copy/paste of the panda-play source codes, etc.

I've not yet included Scala 3 builds - we'll need to get to the new play versions before we can do that upgrade, but I'll give it a try soon anyway - looks less trivial since there's dependencies that aren't available for scala 3 :( but if I understand right we can always use 2.13 builds of panda on scala 3 projects anyway, so I don't feel any need to rush into setting up that build

- [x] Tested with a play 2.9 upgrade https://github.com/guardian/crosswordv2/pull/132
- [ ] Tested with a play 3.0 upgrade (not done but since the 2.9 works I'm pretty confident in the 3.0 too - the only differences are pekko over akka (not referenced directly in panda), and the dependency group ID)


davidfurey (approved):
lgtm

PR #123, opened 2023-10-31
Changes required to fully and correctly release 1.4.0

andrew-nowak:
@guardian/digital-cms

I made a couple of mistakes in the previous PR - to correctly get sbt-release to release a multiproject with differing cross scala configurations, we need to set releaseCrossBuild off (confusing! but gives control over crossbuilding to sbt, instead of sbt-release, and sbt handles it better); and include the sonatypeReleaseSettings on the _root_ project, _and_ sonatype won't publish without a homepage set.

Once all of that is done, the only and only correct way to run a release is `sbt clean release` (and NOT `sbt clean +release`, `sbt clean "release cross"` or any other invocation which refers to crossbuilding!)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -129,7 +129,8 @@ lazy val root = Project("pan-domain-auth-root", file(".")).aggregate(
   panDomainAuthPlay_2_9,
   panDomainAuthPlay_3_0,
   exampleApp
-).settings(
+).settings(sonatypeReleaseSettings)
+ .settings(
   organization := "com.gu",
   publishArtifact := false,
   publish / skip := true,
--- a/build.sbt
+++ b/build.sbt
@@ -11,6 +11,22 @@ val scala213 = "2.13.8"
 
 ThisBuild / scalaVersion := scala213
 
+val checkRunCorrectly = ReleaseStep(action = st => {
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -29,13 +28,14 @@ val sonatypeReleaseSettings = {
...
     )),
+    homepage := Some(url("https://github.com/guardian/pan-domain-authentication")),
     developers := List(Developer(
       id = "GuardianEdTools",
       name = "Guardian Editorial Tools",
       email = "digitalcms.dev@theguardian.com",
       url = url("https://github.com/orgs/guardian/teams/digital-cms")
     )),
-    releaseCrossBuild := true,
+    releaseCrossBuild := false,
```
jonathonherbert, line 38: Appreciate it's in VCS, but wonder if it's worth adding a comment here – as you mention this is a bit weird!

jonathonherbert (approved):

build.sbt
andrew-nowak: done; I also went mildly overboard and included a new release step which checks the user hasn't tried to invoke sbt with a cross-build instruction while releasing - which caused me some time loss while figuring out the correct incantation

jonathonherbert: Nice. Back of my mind: moving to autorelease for these libraries would likely reduce confusion too!

PR #124, opened 2023-11-20
Pull panda-hmac into repo

andrew-nowak:
@guardian/digital-cms

Pull the contents of the [panda-hmac](https://github.com/guardian/panda-hmac) repo into this one, allowing us to release all panda-related modules simultaneously and reducing repo fragmentation.


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "1.4.1-SNAPSHOT"
+version in ThisBuild := "3.0.0-SNAPSHOT"
```
andrew-nowak, line 1: big bump but IMO, panda-hmac should be on the same version numbers as the other modules, but currently panda-hmac is on 2.2.0 and the rest of panda is on 1.4.0.
This could be 2.3.0 but it seems weird to me that most panda modules would not only jump to a new major version, but not even have a x.0.0 release? maybe i'm overthinking it though

jonathonherbert (approved):
👍 👍 

IIUC it's only the play actions that have a dependency on panda – wonder if this repo will eventually be `gu-authentication` 🤔

andrew-nowak:
are you referring to hmac-headers? yeah I'm leaving that repo where it is since it's nicely independent of the rest of the panda ecosystem

PR #133, opened 2024-01-17
Retry when discovery document fetcher fails

andrew-nowak:
@guardian/digital-cms

Motivated by seeing an example where DNS resolution failed for the discovery document, and that failure was held for the rest of the instance's lifetime. If when asking for the discovery document future, we see that the fetch has failed; replace the future with a fresh fetch.

### DIFF
--- a/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
+++ b/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
@@ -16,15 +16,20 @@ import scala.language.postfixOps
 class OAuthException(val message: String, val throwable: Throwable = null) extends Exception(message, throwable)
 
 class OAuth(config: OAuthSettings, system: String, redirectUrl: String)(implicit context: ExecutionContext, ws: WSClient) {
-  private val discoveryDocumentHolder: AtomicReference[Future[DiscoveryDocument]] =
-    new AtomicReference[Future[DiscoveryDocument]](fetchDiscoveryDocument)
 
-  private def fetchDiscoveryDocument: Future[DiscoveryDocument] =
-    ws.url(config.discoveryDocumentUrl).get().map(r => DiscoveryDocument.fromJson(r.json))
+  private val discoveryDocumentHolder: AtomicReference[Future[DiscoveryDocument]] =
+    new AtomicReference[Future[DiscoveryDocument]](fetchDiscoveryDocument())
+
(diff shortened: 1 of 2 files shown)


pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
```diff
@@ -1,30 +1,31 @@
...
-    else {
-      val discoveryDocumentFuture = ws.url(config.discoveryDocumentUrl).get().map(r => DiscoveryDocument.fromJson(r.json))
-      discoveryDocumentHolder = Some(discoveryDocumentFuture)
-      discoveryDocumentFuture
-    }
+  private def fetchDiscoveryDocument: Future[DiscoveryDocument] =
+    ws.url(config.discoveryDocumentUrl).get().map(r => DiscoveryDocument.fromJson(r.json))
+
+  private def discoveryDocument: Future[DiscoveryDocument] =
+    discoveryDocumentHolder.updateAndGet(ddf =>
```
davidfurey, line 26: `ddf` isn't a great name but I can't think of anything better.

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
```diff
@@ -45,7 +46,7 @@ class OAuth(config: OAuthSettings, system: String, redirectUrl: String) {
   }
 
   def redirectToOAuthProvider(antiForgeryToken: String, email: Option[String] = None)
-                      (implicit context: ExecutionContext, request: RequestHeader, ws: WSClient): Future[Result] = {
+                      (implicit context: ExecutionContext): Future[Result] = {
```
davidfurey, line 49: It isn't immediately obvious to me why you have moved these implicit parameters to the class level

davidfurey (commented):
Approach seems reasonable, just a bit confused about the change to the implicits

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
andrew-nowak: They're not strictly necessary, but it made life a bunch easier to move the implicits for the `discoveryDocument` method up to be shared throughout the class, so that I can make the initial discovery doc request at construction. Once they're up at the class level, these implicits become unnecessary, because the values used in the construction of the OAuth class are also the values that are passed to `redirectToOAuthProvider`.

davidfurey (approved):

PR #136, opened 2024-04-17
update to latest scalas (2.12 and 2.13) to avoid binary incompatibility on 2.13.9

andrew-nowak:
also apply recommended new compiler warning fixes from later version

see <https://github.com/scala/bug/issues/12650> for 2.13.9 binary incompatibility


twrichards (approved):
Changes LGTM

PR #137, opened 2024-04-23
expose an `authenticateRequest` function

twrichards:
...on `AuthAction` & `APIAuthAction` which can be called with a `RequestHeader` allowing it to be used in play filters for example.

frontend uses play filters to ensure all of `preview` and `admin` endpoints are wrapped in auth... in https://github.com/guardian/frontend/pull/27012 we replace plain Google auth with pan-domain-authentication, but filters only deal in `RequestHeader` (understandably) but it would be useful to be able to reuse all of the authentication and redirect logic in the `AuthAction`, fortunately this was relatively straigh-forward (and is solving the problem when published locally).


twrichards:
https://github.com/guardian/frontend/pull/27012 is all working nicely locally, so need to release this library version asap before I can test frontend changes in CODE (and release)

andrew-nowak (approved):

bryophyta (approved):

PR #146, opened 2024-06-27
Update docs to link to Emergency Panda key-rotation Guide

rtyley:
@andrew-nowak & I did a walkthrough in the CODE environment of the process necessary to do an Emergency rotation of the Panda cookie-signing key, and prepared a [full guide to the steps](https://docs.google.com/document/d/1haVnQ9D8zNYUU-fOfkudPC1WpPGrlelLygd8V7xb3eQ/edit?usp=sharing) (incidentally, I've placed this doc into the P&E shared drive to ensure it remains accessible).

This PR just links to that Google document!

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,8 @@ publicKey=example_key
 
 ### Generating Keys
 
+**Guardian Devs**: See the [Emergency Panda key-rotation Guide](https://docs.google.com/document/d/1haVnQ9D8zNYUU-fOfkudPC1WpPGrlelLygd8V7xb3eQ/edit?usp=sharing).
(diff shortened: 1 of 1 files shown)


dblatcher (approved):
Thanks for adding this - not related to the Pr, but I noticed the access to the demo video at the bottom of the linked document was restricted - might be wroth setting view permissions to the same as the document.

rtyley:
> I noticed the access to the demo video at the bottom of the linked document was restricted - might be wroth setting view permissions to the same as the document.

Thanks- I think this is fixed now?!

PR #147, opened 2024-06-27
Use `java.security` classes in preference to string-wrappers

rtyley:
This change deletes the `PublicKey` & `PrivateKey` classes in our `com.gu.pandomainauth` package, which were simple string-wrapper classes that held base64-encoded text representations of cryptographic keys.

The `java.security` classes `PublicKey` & `PrivateKey` already exist and, as they represent fully-decoded keys, don't need to be _repeatedly_ decoded from strings the way we had to with our string-wrapper classes. Consequently, the `signData()` &`verifySignature()` methods on `Crypto` are now a bit simpler and more efficient.

We can also use the `java.security.KeyPair` class to represent the pair of public & private keys.

### DIFF
--- a/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/PanDomainAuthSettings.scala
+++ b/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/PanDomainAuthSettings.scala
@@ -1,10 +1,11 @@
 package com.gu.pandomainauth.model
 
-import com.gu.pandomainauth.{PrivateKey, PublicKey}
+import com.gu.pandomainauth.service.Crypto
+
+import java.security.KeyPair
 
 case class PanDomainAuthSettings(
-  publicKey: PublicKey,
-  privateKey: PrivateKey,
+  signingKeyPair: KeyPair,
   cookieSettings: CookieSettings,
   oAuthSettings: OAuthSettings,
   google2FAGroupSettings: Option[Google2FAGroupSettings]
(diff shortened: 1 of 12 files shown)


pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/Crypto.scala
```diff
@@ -25,19 +24,25 @@ object Crypto {
...
 
     rsa.update(data)
     rsa.sign()
   }
 
   def verifySignature(data: Array[Byte], signature: Array[Byte], pubKey: PublicKey) : Boolean = {
     val rsa = Signature.getInstance(signatureAlgorithm, "BC")
-    val publicKey = keyFactory.generatePublic(new X509EncodedKeySpec(decodeBase64(pubKey.key)))
-    rsa.initVerify(publicKey)
+    rsa.initVerify(pubKey)
```
rtyley, line 35: The `signData()` &`verifySignature()` methods on `Crypto` no-longer need to repeatedly decode the base64-encoded strings, and so now are a bit simpler and more efficient.

pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/model/PanDomainAuthSettings.scala
```diff
@@ -1,10 +1,11 @@
...
 
-import com.gu.pandomainauth.{PrivateKey, PublicKey}
+import com.gu.pandomainauth.service.Crypto
+
+import java.security.KeyPair
 
 case class PanDomainAuthSettings(
-  publicKey: PublicKey,
-  privateKey: PrivateKey,
+  signingKeyPair: KeyPair,
```
rtyley, line 8: The signature of the `PanDomainAuthSettings` case class has changed here, but thankfully this won't affect consumers of this library, as they all use the companion `apply()` method that accepts `settingMap: Map[String, String]`.

davidfurey (approved):
Looks good to me

PR #150, opened 2024-07-30
Support accepting multiple public keys

rtyley:
This PR introduces a way for Panda keys to be manually rotated without disruption to users, and is a stepping stone towards the later goal of automatically rotating keys on a regular schedule. Consequently, some of the work here is temporary (like the `CryptoConfForRotation` script) and is intended for later replacement.

### Changes to `.settings` & `.settings.public` config files

Panda's settings for signing & verification keys have been extended, while keeping the config files backwards-compatible with old clients. The existing configuration keys remain unchanged, and can still only be used at most once in a config file:

- `privateKey`
- `publicKey`

...semantically, those settings now relate specifically to the _active_ key pair (ie the one that is _currently_ doing cookie-signing). In addition, there is now a pattern for settings that list _additional_ public keys which will _also_ be accepted:

- `alsoAccept.xxx.publicKey` - where `xxx` can be any string that's valid in a property key identifier, allowing multiple entries like this to be in a config file

### Generating a new key pair

The script used to generate a new cryptographic key pair has been replaced:

- **OLD**: `generateKeyPair.sh` bash script, using [`openssl`](https://wiki.openssl.org/index.php/Command_Line_Utilities) to generate the new key pair.
- **NEW** : `CryptoConfForRotation` Scala code, executable from the `sbt` prompt, using [`java.security.KeyPairGenerator`](https://docs.oracle.com/javase/8/docs/api/java/security/KeyPairGenerator.html) to generate a new key pair. The new script also generates the 3 phases of config required for a smooth key rotation.

### Performing a smooth key rotation

To perform a smooth key rotation, without disruption to users, there needs to be 3 separate updates to each config file in S3, each with a delay between each step. The 4 phases can be seen in the example `.settings` files in [src/test/resources/crypto-conf-rotation-example](https://github.com/guardian/pan-domain-authentication/tree/support-accepting-multiple-public-keys/pan-domain-auth-verification/src/test/resources/crypto-conf-rotation-example):

0. [legacy.settings](https://github.com/guardian/pan-domain-authentication/blob/support-accepting-multiple-public-keys/pan-domain-auth-verification/src/test/resources/crypto-conf-rotation-example/0.legacy.settings) - the settings before a rotation, with just a single active key-pair, defined by `privateKey` & `publicKey`
1. [rotation-upcoming.settings](https://github.com/guardian/pan-domain-authentication/blob/support-accepting-multiple-public-keys/pan-domain-auth-verification/src/test/resources/crypto-conf-rotation-example/1.rotation-upcoming.settings) - the same settings, but now also with a `alsoAccept.0.publicKey` entry that specifies the public key of the _upcoming_ active key-pair. Once this update has been applied to S3, there should be a delay of at least **1 minute** before the next phase, to ensure that _all_ Panda services are ready to accept a cookie signed with the upcoming active key-pair.
2. [rotation-in-progress.settings](https://github.com/guardian/pan-domain-authentication/blob/support-accepting-multiple-public-keys/pan-domain-auth-verification/src/test/resources/crypto-conf-rotation-example/2.rotation-in-progress.settings) - the active key-pair (`privateKey` & `publicKey`) is now the _new_ key-pair, and `alsoAccept.0.publicKey` is now set to the public key of the _old_ active key-pair. Once this update has been applied, there should be a delay of at least **1 hour**, to ensure all users have regenerated their cookies.
3. [rotation-complete.settings](https://github.com/guardian/pan-domain-authentication/blob/support-accepting-multiple-public-keys/pan-domain-auth-verification/src/test/resources/crypto-conf-rotation-example/3.rotation-complete.settings) - finally, we remove `alsoAccept.0.publicKey`, and only the new active key-pair is accepted.

Instructions in the `ReadMe.md` have been updated to reflect this.

### Testing

Need to update these PRs (originally created for https://github.com/guardian/pan-domain-authentication/pull/151):

* https://github.com/guardian/atom-workshop/pull/361
* https://github.com/guardian/login.gutools/pull/81
* https://github.com/guardian/s3-upload/pull/58

### Supporting PRs

This PR sits on top of:

* https://github.com/guardian/pan-domain-authentication/pull/151
* https://github.com/guardian/pan-domain-authentication/pull/152
* https://github.com/guardian/pan-domain-authentication/pull/155


pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CryptoConf.scala
```diff
@@ -42,4 +74,33 @@ object CryptoConf {
     def privateKeyFor(base64Key: String): SettingsResult[PrivateKey] =
       keyFor(base64Key, new PKCS8EncodedKeySpec(_), _.generatePrivate)
   }
+
+  object Change {
```
bryophyta, line 78: I found it quite difficult to parse the logic in this function. I feel as though it could be easier to read if it was made a little less concise, e.g. something like:

```
      if (newConf == oldConf) {
        None
      } else {
        val maybeActiveKey = if (newConf.activeKeyPair == oldConf.activeKeyPair) {
          None
        } else {
          Some(
            ActiveKey(
              toleratingOldKey = newConf.alsoAccepted.contains(oldConf.activeKeyPair.publicKey),
              newKeyAlreadyAccepted = oldConf.alsoAccepted.contains(newConf.activeKeyPair.publicKey)
            )
          )
        }
        Some(Change(
          activeKey = maybeActiveKey,
          SeqDiff.compare(oldConf.alsoAccepted, newConf.alsoAccepted)
        ))
      }
```

However, even once I'd parsed the logic I found the `Change` and `ActiveKey` abstractions a bit difficult to understand. Possibly it could be made a little easier by adjusting some naming, but if I understand correctly these classes are only used for generating logging messages, so I wonder if it'd be worth considering something a bit less elegant but perhaps more digestible, e.g.

```
val oldAndNewKeysAreTheSame = ???
val oldKeyIsDifferentButStillTolerated = ???
val newKeyIsAlreadyAccepted = ???

val (added, removed) = SeqDiff.compare(???)

val summary = ???
```

Maybe my suggestion above is a bit _too_ plodding, but I wonder if there's a happy middle ground available?

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
```diff
@@ -21,15 +22,24 @@ class PublicSettings(loader: Settings.Loader, scheduler: ScheduledExecutorServic
     new Settings.Loader(S3BucketLoader.forAwsSdkV1(s3Client, bucketName), settingsFileKey), scheduler
   )
 
-  private val settingsRefresher = new Settings.Refresher[PublicKey](
+  private val settingsRefresher = new Settings.Refresher[Verification](
     loader,
-    CryptoConf.SettingsReader(_).activePublicKey,
+    CryptoConf.SettingsReader(_).verificationConf,
+    (o, n) => {
+//      for (change <- CryptoConf.Change.compare(o, n)) {
```
bryophyta, line 29: are these lines meant to be commented out?

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/Crypto.scala
```diff
@@ -14,22 +14,21 @@ object Crypto {
    *
    * Note: you only need to pass the key ie the blob of base64 between the start and end markers in the pem file.
    */
-
   Security.addProvider(new BouncyCastleProvider())
 
-  val signatureAlgorithm: String = "SHA256withRSA"
   val keyFactory = KeyFactory.getInstance("RSA")
+  private def signatureInstance() = Signature.getInstance("SHA256withRSA", "BC")
```
bryophyta, line 20: 👍

bryophyta (commented):
this looks to me like it should work as intended, but i've made a couple of comments where i thought that the code could be easier for a newcomer to read

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CryptoConf.scala
bryophyta: thanks very much for the detailed explanation!

```scala
activeKey = Option.when(newConf.activeKeyPair != oldConf.activeKeyPair)(ActiveKey(
      toleratingOldKey = newConf.acceptsActiveKeyFrom(oldConf),
      newKeyAlreadyAccepted = oldConf.acceptsActiveKeyFrom(newConf)
)),
```

this alternative feels much more readable and explicit to me!

rtyley:
This has been tested with two apps running Panda at 8e972972:

* https://github.com/guardian/atom-workshop/pull/361
* https://github.com/guardian/login.gutools/pull/81

In [this video](https://drive.google.com/file/d/1s8hVeJsu88Oj2SzYwBX18-5oIMZCb6b1/view?usp=sharing), you can see me perform all 3 stages of key rotation in the CODE environment.

The logs for this are [here](https://logs.gutools.co.uk/s/editorial-tools/goto/e099e940-71e4-11ef-9ff8-9912aeb0ea2d) (note that a good log filter is to filter for the `logger_name` of `com.gu.pandomainauth.Settings$Refresher`):

image

README.md
```diff
@@ -155,20 +155,26 @@ publicKey=example_key
 
 * **publicKey** - this is the public key used to verify the cookie
 
-### Generating Keys
+### Rotating Keys
 
 **Guardian Devs**: See the [Emergency Panda key-rotation Guide](https://docs.google.com/document/d/1haVnQ9D8zNYUU-fOfkudPC1WpPGrlelLygd8V7xb3eQ/edit?usp=sharing).
```
bryophyta, line 160: would it make sense to split out emergency rotation vs. routine rotation under separate headings?

bryophyta (approved):
We've discussed a couple of iterations on this in some depth and I've read through the most recent version as well. LGTM!

README.md
rtyley: Actually, I think the word 'emergency' can be removed from a few parts of the documentation now, as, with this PR, key-rotation is no longer an emergency-only procedure! So I've added 56b456ecd952f5400499f3bbf14575c269b08375 to take out the word 'emergency' here, and once this PR is merged I'll be updating [_Emergency Panda key rotation instructions_](https://docs.google.com/document/d/1haVnQ9D8zNYUU-fOfkudPC1WpPGrlelLygd8V7xb3eQ/edit) to be titled just _Guardian Panda key rotation instructions_, and of course updating the instructions within too.

rtyley:
Slightly annoyingly, the merging of this PR went slightly wrong - because this PR was sitting on top of https://github.com/guardian/pan-domain-authentication/pull/155, and that PR had been merged to `main` only seconds earlier at 12:50pm, when I merged _this_ PR moments later, it was merged into the now-redundant `fix-creation-of-superfluous-PanDomainAuthSettingsRefresher-instances` branch of #155, rather than `main`. I didn't see this, and kicked off a release, so https://github.com/guardian/pan-domain-authentication/releases/tag/v6.0.0 doesn't contain this change! I'll have to merge that branch into `main` now and do another release...

PR #151, opened 2024-08-07
Refactor Settings loading-&-parsing

rtyley:
This is a precursor to PR https://github.com/guardian/pan-domain-authentication/pull/150 (_"Support accepting multiple public keys"_). It helps with that PR because it consolidates the loading-&-parsing code for configuration settings (previously duplicated across the code for the `.settings` and `.settings.public` files), so that when we add support for configuring multiple public keys, we don't need to repeat the implementation.

* **New type alias `SettingsResult`**: rather than throw exceptions, communicate `SettingsFailure`s with return values of `SettingsResult`.
* **New class `Settings.Loader`**: takes responsibility for getting a settings `Map[String, String]` from S3, previously implemented in both `PanDomainAuthSettingsRefresher` and `PublicSettings`:
  https://github.com/guardian/pan-domain-authentication/blob/59c713202bfe9b301db4d70dda0480b73c89fba0/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala#L43
  https://github.com/guardian/pan-domain-authentication/blob/59c713202bfe9b301db4d70dda0480b73c89fba0/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala#L66
* **New class `Settings.Refresher`**: polls S3 for settings updates, extracting common implementation from `PanDomainAuthSettingsRefresher` & `PublicSettings`.

### Testing

* https://github.com/guardian/atom-workshop/pull/361
* https://github.com/guardian/login.gutools/pull/81
* https://github.com/guardian/s3-upload/pull/58

### See also

* https://github.com/guardian/pan-domain-authentication/pull/150

### DIFF
--- a/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
+++ b/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
@@ -1,51 +1,35 @@
 package com.gu.pandomainauth
 
-import java.util.concurrent.atomic.AtomicReference
-import java.util.concurrent.{Executors, ScheduledExecutorService, TimeUnit}
 import com.amazonaws.services.s3.AmazonS3
-import com.gu.pandomainauth.PublicSettings.validateAndParseKeyText
-import com.gu.pandomainauth.service.Crypto
-import org.slf4j.LoggerFactory
+import com.gu.pandomainauth.Settings.Loader
+import com.gu.pandomainauth.SettingsFailure.SettingsResult
+import com.gu.pandomainauth.service.CryptoConf
 
 import java.security.PublicKey
-import java.util.regex.Pattern
-import scala.concurrent.ExecutionContext
+import java.util.concurrent.Executors.newScheduledThreadPool
+import java.util.concurrent.{Executors, ScheduledExecutorService}
 import scala.concurrent.duration._
 
 /**
  * Class that contains the static public settings and includes mechanism for fetching the public key. Once you have an
  * instance, call the `start()` method to load the public data.
-  *
-  * @param settingsFileKey the settings file for the domain in the S3 bucket (eg local.dev.gutools.co.uk.public.settings)
-  * @param bucketName      the name of the S3 bucket (eg pan-domain-auth-settings)
-  * @param s3Client        the AWS S3 client that will be used to download the settings from the bucket
-  * @param scheduler       optional scheduler that will be used to run the code that updates the bucket
  */
-class PublicSettings(settingsFileKey: String, bucketName: String, s3Client: AmazonS3,
-                     scheduler: ScheduledExecutorService = Executors.newScheduledThreadPool(1)) {
+class PublicSettings(loader: Settings.Loader, scheduler: ScheduledExecutorService) {
 
-  private val agent = new AtomicReference[Option[PublicKey]](None)
+  def this(settingsFileKey: String, bucketName: String, s3Client: AmazonS3,
+    scheduler: ScheduledExecutorService = Executors.newScheduledThreadPool(1)) = this(
+    new Settings.Loader(S3BucketLoader.forAwsSdkV1(s3Client, bucketName), settingsFileKey), scheduler
+  )
 
-  private val logger = LoggerFactory.getLogger(this.getClass)
-  implicit private val executionContext: ExecutionContext = ExecutionContext.fromExecutor(scheduler)
+  private val settingsRefresher = new Settings.Refresher[PublicKey](
+    loader,
+    CryptoConf.SettingsReader(_).activePublicKey,
+    scheduler
+  )
 
-  def start(interval: FiniteDuration = 60.seconds): Unit = {
-    scheduler.scheduleAtFixedRate(() => refresh(), 0, interval.toMillis, TimeUnit.MILLISECONDS)
-  }
+  def start(interval: FiniteDuration = 60.seconds): Unit = settingsRefresher.start(interval.toMinutes.toInt)
 
-  def refresh(): Unit = {
-    PublicSettings.getPublicKey(settingsFileKey, bucketName, s3Client) match {
-      case Right(publicKey) =>
-        agent.set(Some(publicKey))
-        logger.debug("Successfully updated pan-domain public settings")
-
-      case Left(err) =>
-        logger.error("Failed to update pan-domain public settings")
-        Settings.logError(err, logger)
-    }
-  }
-
-  def publicKey: Option[PublicKey] = agent.get()
+  def publicKey: PublicKey = settingsRefresher.get()
 }
 
 /**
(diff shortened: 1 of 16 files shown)


pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/Crypto.scala
```diff
@@ -37,12 +35,4 @@ object Crypto {
     rsa.update(data)
     rsa.verify(signature)
   }
-
-  def publicKeyFor(base64EncodedKey: String): PublicKey =
-    keyFactory.generatePublic(new X509EncodedKeySpec(decodeBase64(base64EncodedKey)))
-  def privateKeyFor(base64EncodedKey: String): PrivateKey =
-    keyFactory.generatePrivate(new PKCS8EncodedKeySpec(decodeBase64(base64EncodedKey)))
```
rtyley, line 44: These methods have moved to `CryptoConf.SettingsReader`.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PublicSettings.scala
```diff
@@ -1,76 +1,50 @@
...
-  * @param settingsFileKey the settings file for the domain in the S3 bucket (eg local.dev.gutools.co.uk.public.settings)
-  * @param bucketName      the name of the S3 bucket (eg pan-domain-auth-settings)
-  * @param s3Client        the AWS S3 client that will be used to download the settings from the bucket
-  * @param scheduler       optional scheduler that will be used to run the code that updates the bucket
  */
-class PublicSettings(settingsFileKey: String, bucketName: String, s3Client: AmazonS3,
-                     scheduler: ScheduledExecutorService = Executors.newScheduledThreadPool(1)) {
+class PublicSettings(loader: Settings.Loader, scheduler: ScheduledExecutorService) {
 
-  private val agent = new AtomicReference[Option[PublicKey]](None)
```
bryophyta, line 27: I'm sure this change is legit, but I don't think I'm clear in my own mind about how either:
1. `AtomicReference` is not needed here, or else
2. it's removal has been compensated for elsewhere

bryophyta: ah, I think I've tracked it down -- the store has been moved to the shared `Settings` class https://github.com/guardian/pan-domain-authentication/pull/151/files#diff-91c51062f921c3d5c266a62c85f90eb5e6afce3eeeea0dbf299e9a7f8320988c

rtyley: That's right! The code is now in `Settings.Refresher`, here:

https://github.com/guardian/pan-domain-authentication/blob/c4835010a87c1ae854c3e8ca4dadf086b5ceb6a4/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/Settings.scala#L82-L85

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/Settings.scala
```diff
@@ -1,66 +1,105 @@
...
 
-import com.amazonaws.services.s3.AmazonS3
-import com.amazonaws.util.IOUtils
-import org.slf4j.Logger
+sealed trait SettingsFailure {
+  val description: String
 
-import scala.util.control.NonFatal
-import scala.jdk.CollectionConverters._
+  def logError(logger: Logger): Unit = logger.error(description)
```
bryophyta, line 18: this seems like a nice abstraction 👍

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/Settings.scala
```diff
@@ -1,66 +1,105 @@
...
-      new IllegalStateException("Unable to parse public key", cause)
+    def loadAndParseSettings(): SettingsResult[A] =
+      loader.loadAndParseSettingsMap().flatMap(settingsParser)
 
-    case PublicKeyFormatFailure =>
-      new IllegalStateException("Public key does not match expected format")
+    private def refresh(): Unit = loadAndParseSettings() match {
+      case Right(newSettings) =>
+        // logger.debug(s"Updated pan-domain settings for $domain")
+        val oldSettings = store.getAndSet(newSettings)
```
bryophyta, line 97: it doesn't look like we're doing anything with `oldSettings` here?

(and should the commented-out logger line above be removed?)

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CryptoConf.scala
```diff
@@ -0,0 +1,44 @@
...
+    val activePublicKey: SettingsResult[PublicKey] = setting("publicKey").flatMap(publicKeyFor)
+
+    def activeKeyPair: SettingsResult[KeyPair] = for {
+      publicKey <- activePublicKey
+      privateKey <- setting("privateKey").flatMap(privateKeyFor)
+    } yield KeyPair(publicKey, privateKey)
+  }
+
+  object SettingsReader {
+    def publicKeyFor(data: Array[Byte]) = keyFactory.generatePublic(new X509EncodedKeySpec(data))
```
bryophyta, line 28: there are two public definitions of `publicKeyFor` (and `privateKeyFor`) in this object -- just wanted to check that both are meant to be exposed publicly? What's the expected use-case for a consumer to use `def publicKeyFor(data: Array[Byte])` as opposed to the `publicKeyFor(base64EncodedKey: String)` method?

pan-domain-auth-verification/src/test/scala/com/gu/pandomainauth/PublicSettingsTest.scala
```diff
@@ -1,38 +1,11 @@
...
-import com.gu.pandomainauth.service.{Crypto, TestKeys}
-import org.scalatest.concurrent.ScalaFutures
-import org.scalatest.freespec.AnyFreeSpec
 import org.scalatest.EitherValues
+import org.scalatest.freespec.AnyFreeSpec
 import org.scalatest.matchers.should.Matchers
 
 
-class PublicSettingsTest extends AnyFreeSpec with Matchers with EitherValues with ScalaFutures {
-  "validateKey" - {
```
bryophyta, line 12: test cases have moved to `CryptoConfTest` to reflect the refactor in the code above 👍

bryophyta (approved):
We've been over a lot of this in conversation together, and I've read through the code separately too. This looks good to me 👍

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/Settings.scala
rtyley: Good point - addressed with 412c1d6f4efa42c68f6f25cec73a74d850b604ad!

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CryptoConf.scala
rtyley: Yeah, good point - the overloading of method names was a bit confusing - addressed with 4c87946a113b73f23e9e9b683082bdaeb49a5e2d!

Issue #149, opened 2024-07-25
Drop Scala 2.12 support


rtyley:
Dropping Scala 2.12 support would allow us to use all the good Scala 2.13 features, like `LazyList`, etc!

### Current blockers

There are 4 non-archived repos that use Panda, still on Scala 2.12:

- [ ] [guardian/grid](https://github.com/guardian/grid/blob/b28b1b272da95e7e3c2901aca0367d03be6c43c2/build.sbt#L18)
- [ ] [guardian/flexible-content](https://github.com/guardian/flexible-content/blob/8e202c24fdc34028e364f5050607092e80f87538/build.sbt#L13)
- [ ] [guardian/tagmanager](https://github.com/guardian/tagmanager/blob/5d11f5decf5c87ec588e69d4ec0f10b2bdf1ed5d/build.sbt#L7)
- [ ] [presence-indicator](https://github.com/guardian/presence-indicator/blob/5388a6b30f0d6d3cef393f1e84d7abeecb35daee/build.sbt#L5)

emdash-ie:
I think grid may also be on scala 2.13 now, as of https://github.com/guardian/grid/pull/4361?

rtyley:
> I think grid may also be on scala 2.13 now, as of [guardian/grid#4361](https://github.com/guardian/grid/pull/4361)?

Amazing! I didn't know about that, awesome!

PR #152, opened 2024-08-07
Refactor Cookie generation-&-parsing

rtyley:
This is a precursor to PR https://github.com/guardian/pan-domain-authentication/pull/150 (_"Support accepting multiple public keys"_). It helps with that PR because we're going to be changing the cookie processing code to validate against a _range_ of possible acceptable keys, and if a failure occurs in the reading of a cookie, we want to be clear about whether that happens to be due to the particular key we're attempting to validate against, or whether the cookie text itself is malformatted.

* **New class `CookiePayload`**: an optimal representation of cookie data for checking signature-validity against *multiple* public keys. The `CookiePayload` class represents the binary data caught after parsing-and-base64-decoding the cookie text value, ready for signature verification.
* **New type alias `CookieResult`**: rather than throw exceptions, communicate `CookieIntegrityFailure`s with return values of `CookieResult`.

### DIFF
--- a/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookiePayload.scala
+++ b/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookiePayload.scala
@@ -1,0 +1,56 @@
+package com.gu.pandomainauth.service
+
+import com.gu.pandomainauth.service.CookiePayload.encodeBase64
+import org.apache.commons.codec.binary.Base64
+import org.apache.commons.codec.binary.Base64.isBase64
+
+import java.nio.charset.StandardCharsets.UTF_8
+import java.security.{PrivateKey, PublicKey}
+import scala.util.matching.Regex
+
+/**
+ * A representation of the underlying binary data (both payload & signature) in a Panda cookie.
+ *
+ * If an instance has been parsed from a cookie's text value, the existence of the instance
+ * *does not* imply that the signature has been verified. It only means that the cookie text was
+ * correctly formatted (two Base64 strings separated by '.').
+ *
+ * `CookiePayload` is designed to be the optimal representation of cookie data for checking
+ * signature-validity against *multiple* possible accepted public keys. It's a bridge between
+ * these two contexts:
+ *
+ * * cookie text: the raw cookie value - two Base64-encoded strings (payload & signature), separated by '.'
+ * * payload text: in Panda, a string representation of `AuthenticatedUser`
+ *
+ * To make those transformations, you need either a public or private key:
+ *
+ * * payload text -> cookie text: uses a *private* key to generate the signature
+ * * cookie text -> payload text: uses a *public* key to verify the signature
+ */
+case class CookiePayload(payloadBytes: Array[Byte], sig: Array[Byte]) {
+  def payloadTextVerifiedSignedWith(publicKey: PublicKey): Option[String] =
+    if (Crypto.verifySignature(payloadBytes, sig, publicKey)) Some(new String(payloadBytes, UTF_8)) else None
+
+  lazy val asCookieText: String = s"${encodeBase64(payloadBytes)}.${encodeBase64(sig)}"
+}
+
+object CookiePayload {
+  private val CookieRegEx: Regex = "^([\\w\\W]*)\\.([\\w\\W]*)$".r
+
+  private def encodeBase64(data: Array[Byte]): String = new String(Base64.encodeBase64(data), UTF_8)
+  private def decodeBase64(text: String): Array[Byte] = Base64.decodeBase64(text.getBytes(UTF_8))
+
+  /**
+   * @return `None` if the cookie text is incorrectly formatted (ie not "abc.xyz", with a '.' separator)
+   */
+  def parse(cookieText: String): Option[CookiePayload] = cookieText match {
+    case CookieRegEx(data, sig) if isBase64(data) && isBase64(sig) =>
+      Some(CookiePayload(decodeBase64(data), decodeBase64(sig)))
+    case _ => None
+  }
+
+  def generateForPayloadText(payloadText: String, privateKey: PrivateKey): CookiePayload = {
+    val data = payloadText.getBytes(UTF_8)
+    CookiePayload(data, Crypto.signData(data, privateKey))
+  }
+}
--- a/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
+++ b/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
@@ -51,9 +51,7 @@ object CookieUtils {
 
   def generateCookieData(authUser: AuthenticatedUser, prvKey: PrivateKey): String =
     CookiePayload.generateForPayloadText(serializeAuthenticatedUser(authUser), prvKey).asCookieText
-
-  // We would quite like to know, if a user is using an old (but accepted) key, *who* that user is- or to put it another
-  // way, give me the authenticated user, and tell me which key they're using
+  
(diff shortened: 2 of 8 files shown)


pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookiePayload.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import java.nio.charset.StandardCharsets.UTF_8
+import java.security.{PrivateKey, PublicKey}
+import scala.util.matching.Regex
+
+/**
+ * A representation of the underlying binary data (both payload & signature) in a Panda cookie.
+ *
+ * If an instance has been parsed from a cookie's text value, the existence of the instance
+ * *does not* imply that the signature has been verified. It only means that the cookie text was
```
bryophyta, line 15: this abstraction seems very useful to me, though I wonder if there's a way to make it more obvious from the name of the case class that having a `CookiePayload` doesn't imply that its signature's been verified?

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
```diff
@@ -16,48 +25,39 @@ object CookieUtils {
...
+      email <- data.get("email")
+      system <- data.get("system")
+      authedIn <- data.get("authedIn")
+      expires <- data.get("expires")
+      multifactor <- data.get("multifactor")
+    } yield AuthenticatedUser(
+      user = User(firstName, lastName, email, data.get("avatarUrl")),
+      authenticatingSystem = system,
+      authenticatedIn = Set(authedIn.split(",").toSeq :_*),
+      expires = expires.toLong,
```
bryophyta, line 47: The return type of `deserializeAuthenticatedUser` has been changed to `Option[AuthenticatedUser]` in this branch. Is the intention that all foreseeable errors should therefore lead to a `None`, rather than throwing? If so, do we need to change how we're handling the `.toLong` call, given that it can throw?

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
```diff
@@ -16,48 +25,39 @@ object CookieUtils {
...
-            throw new CookieSignatureInvalidException
-          }
-        } catch {
-          case e: SignatureException =>
-            throw new CookieSignatureInvalidException
-        }
-      case _ => throw new CookieParseException
-    }
-  }
+  // We would quite like to know, if a user is using an old (but accepted) key, *who* that user is- or to put it another
```
bryophyta, line 55: I don't think I follow how this code does what the comment says..

From what I can see it'll either return an `AuthenticatedUser`, or a `SignatureNotValid` (or one of the other failure types), but none of these include the key that the user was using.. Am I missing something?

pan-domain-auth-verification/src/test/scala/com/gu/pandomainauth/service/CookiePayloadTest.scala
```diff
@@ -0,0 +1,37 @@
+package com.gu.pandomainauth.service
+
+import org.scalatest.OptionValues
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.should.Matchers
+import TestKeys._
+
+class CookiePayloadTest extends AnyFreeSpec with Matchers with OptionValues {
```
bryophyta, line 8: it looks to me like these cases provide good cover 👍

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookiePayload.scala
rtyley: Maybe `UnverifiedCookiePayload`? I think bringing in `Unverified` is a bit tricky though, because `CookiePayload` transforms both ways and so changing its name also affects what the code looks like when we _generate_ a cookie:

https://github.com/guardian/pan-domain-authentication/blob/41c7bafe524a1b2bc4a80c5b78c35eb99bfadc1e/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala#L53

...given that at that point we're literally signing it with a private key, it wouldn't feel quite right to me to be referring to it as an _Unverified_ CookiePayload...

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
rtyley: Good point! Addressed in f40aa8d7263952b5499070d3f01b2b4403821939.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
rtyley: Ah, but it [fails](https://github.com/guardian/pan-domain-authentication/actions/runs/10321653497/job/28575045968#step:6:21) because [we have to compile under Scala 2.12](https://github.com/guardian/pan-domain-authentication/issues/149):

```
[error] /home/runner/work/pan-domain-authentication/pan-domain-authentication/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala:41:48: value toLongOption is not a member of String
[error]       expires <- data.get("expires").flatMap(_.toLongOption)
[error]                                                ^
[error] /home/runner/work/pan-domain-authentication/pan-domain-authentication/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala:42:56: value toBooleanOption is not a member of String
[error]       multiFactor <- data.get("multifactor").flatMap(_.toBooleanOption)
[error]                                                        ^
[error] two errors found
```

...I'll make something Scala 2.12 compliant...

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
rtyley: Whoops - ah, but it [fails](https://github.com/guardian/pan-domain-authentication/actions/runs/10321653497/job/28575045968#step:6:21) because [we have to compile under Scala 2.12](https://github.com/guardian/pan-domain-authentication/issues/149):

```
[error] /home/runner/work/pan-domain-authentication/pan-domain-authentication/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala:41:48: value toLongOption is not a member of String
[error]       expires <- data.get("expires").flatMap(_.toLongOption)
[error]                                                ^
[error] /home/runner/work/pan-domain-authentication/pan-domain-authentication/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala:42:56: value toBooleanOption is not a member of String
[error]       multiFactor <- data.get("multifactor").flatMap(_.toBooleanOption)
[error]                                                        ^
[error] two errors found
```

...I'll make something Scala 2.12 compliant...

rtyley: > ...I'll make something Scala 2.12 compliant...

Done with 998db1f097e51ee2d231ce23edcc7536384c4be3!

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookiePayload.scala
bryophyta: ah yes I see your point.. probably it would require having two different data types in that case, which seems like it'd be OTT at this point for a fairly theoretical issue 👍

bryophyta (approved):
this looks good to me 👍

PR #155, opened 2024-09-12
Fix creation of superfluous `PanDomainAuthSettingsRefresher` instances, make `panDomainSettings` a `val`

rtyley:
The field `panDomainSettings` in the `AuthActions` trait was defined as a `def`, which was problematic:

https://github.com/guardian/pan-domain-authentication/blob/65e1f65760a1df773589a2afb1fc2dd09944a89e/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala#L36

Having the field in trait as as a `def` means that implementors of `AuthActions` are able to make the mistake of _also_ defining their implementation of [that field as a `def`](https://github.com/guardian/atom-workshop/blob/2df73bdbb953401bc9480142fa04cee29f577a6b/app/AppComponents.scala#L33), meaning that every read of that field will create _another_ instance of the refresher, [as seen in Atom Workshop](https://github.com/guardian/atom-workshop/pull/361#issuecomment-2346317855). We can see that each EC2 instance has _lots_ of instances of `PanDomainAuthSettingsRefresher`, all logging about the config change:

image

Surprisingly, this seems to have been a problem since at least as far back as https://github.com/guardian/pan-domain-authentication/pull/41 in February 2018 (which was _also_ dealing with a problem where the refresher was instantiated too often!).

## Fix: make the trait field a `val` rather than a `def`

Changing the field type to `val` forces implementers to also use `val` for that field, effectively making it a **singleton**, which is what we want.

Changing the abstract field of a trait to be `val` does open up [another danger due the initialisation order of `val`s](https://docs.scala-lang.org/tutorials/FAQ/initialization-order.html) - the field could end up being evaluated as `null` if the trait immediately evaluates the field during construction.

...consequently, I've made all `val` declarations in the `AuthActions` trait (that evaluate `panDomainSettings` in some way) into `lazy val`s.

### Testing

[Deploying](https://riffraff.gutools.co.uk/deployment/view/05f3707f-20ce-4ffc-98c6-8e467079f001) https://github.com/guardian/atom-workshop/pull/366 at https://github.com/guardian/atom-workshop/commit/b6e8a2cd6ddffd11a775ae8e3770e2bd28a2f026 (using this PR at 61a193ca) to CODE, we see that when we change the Panda settings file in S3, there is, as desired, only **1** instance of `PanDomainAuthSettingsRefresher` per EC2 instance to [log about it](https://logs.gutools.co.uk/goto/89a5e740-7123-11ef-b481-afe258002fe1):

image

### DIFF
--- a/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala
+++ b/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala
@@ -33,10 +33,10 @@ trait AuthActions {
     */
   def wsClient: WSClient
   def controllerComponents: ControllerComponents
-  def panDomainSettings: PanDomainAuthSettingsRefresher
+  val panDomainSettings: PanDomainAuthSettingsRefresher
 
-  private def system: String = panDomainSettings.system
-  private def domain: String = panDomainSettings.domain
+  private lazy val system: String = panDomainSettings.system
+  private lazy val domain: String = panDomainSettings.domain
   private def settings: PanDomainAuthSettings = panDomainSettings.settings
 
   private implicit val ec: ExecutionContext = controllerComponents.executionContext
@@ -82,14 +82,14 @@ trait AuthActions {
     */
   def authCallbackUrl: String
 
-  val OAuth = new OAuth(settings.oAuthSettings, system, authCallbackUrl)(ec, wsClient)
+  lazy val OAuth = new OAuth(settings.oAuthSettings, system, authCallbackUrl)(ec, wsClient)
 
   /**
     * Application name used for initialising Google API clients for directory group checking
     */
-  val applicationName: String = s"pan-domain-authentication-$system"
+  lazy val applicationName: String = s"pan-domain-authentication-$system"
 
-  val multifactorChecker: Option[Google2FAGroupChecker] = settings.google2FAGroupSettings.map {
+  lazy val multifactorChecker: Option[Google2FAGroupChecker] = settings.google2FAGroupSettings.map {
     new Google2FAGroupChecker(_, panDomainSettings.s3BucketLoader, applicationName)
   }
 


pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -33,10 +33,10 @@ trait AuthActions {
     */
   def wsClient: WSClient
   def controllerComponents: ControllerComponents
-  def panDomainSettings: PanDomainAuthSettingsRefresher
+  val panDomainSettings: PanDomainAuthSettingsRefresher
 
-  private def system: String = panDomainSettings.system
-  private def domain: String = panDomainSettings.domain
+  private lazy val system: String = panDomainSettings.system
+  private lazy val domain: String = panDomainSettings.domain
```
rtyley, line 39: Although `system` & `domain` could have remained as `def`s, the truth is that they are constant `val` members of `PanDomainAuthSettingsRefresher`, so they can not change - consequently, they might as well be `lazy val`s (but they must be `lazy`, as otherwise that would immediately evaluate `panDomainSettings` at construction time, which we want to avoid).

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -82,14 +82,14 @@ trait AuthActions {
...
   def authCallbackUrl: String
 
-  val OAuth = new OAuth(settings.oAuthSettings, system, authCallbackUrl)(ec, wsClient)
+  lazy val OAuth = new OAuth(settings.oAuthSettings, system, authCallbackUrl)(ec, wsClient)
 
   /**
     * Application name used for initialising Google API clients for directory group checking
     */
-  val applicationName: String = s"pan-domain-authentication-$system"
+  lazy val applicationName: String = s"pan-domain-authentication-$system"
```
rtyley, line 90: The fields `OAuth` and `applicationName` are both made `lazy` because they both reference `system` - and to evaluate `system`, they need to evaluate `panDomainSettings`, which we want to delay until necessary, due to abstract `val`s in traits [evaluating to `null` during the construction](https://docs.scala-lang.org/tutorials/FAQ/initialization-order.html).

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -82,14 +82,14 @@ trait AuthActions {
...
 
-  val multifactorChecker: Option[Google2FAGroupChecker] = settings.google2FAGroupSettings.map {
+  lazy val multifactorChecker: Option[Google2FAGroupChecker] = settings.google2FAGroupSettings.map {
```
rtyley, line 92: The field `multifactorChecker` is made `lazy` because it references `panDomainSettings` and there is again the risk that it will be `null` if evaluated at the point of construction.

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -33,10 +33,10 @@ trait AuthActions {
     */
   def wsClient: WSClient
   def controllerComponents: ControllerComponents
-  def panDomainSettings: PanDomainAuthSettingsRefresher
+  val panDomainSettings: PanDomainAuthSettingsRefresher
```
rtyley, line 36: This is the core of the fix - we don't want to repeatedly evaluate `panDomainSettings` and get a new one each time - so it shouldn't be a `def`!

bryophyta (approved):
I haven't tested this in CODE but the code looks good to me 👍

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala
```diff
@@ -33,10 +33,10 @@ trait AuthActions {
...
 
-  private def system: String = panDomainSettings.system
-  private def domain: String = panDomainSettings.domain
+  private lazy val system: String = panDomainSettings.system
+  private lazy val domain: String = panDomainSettings.domain
   private def settings: PanDomainAuthSettings = panDomainSettings.settings
```
mxdvl, line 40: Why does this remain a `def`? `settings` seem to be the way to `panDomainSettings` in most instances…

rtyley: Really great question - there are two things that you need to know to understand why we have `def` rather than `val` in `AuthActions` on `def settings: PanDomainAuthSettings = panDomainSettings.settings`.

1. `panDomainSettings` is actually a `PanDomainAuthSettingsRefresher` - it should really be called something like `pandaSettingsRefresher` (tho' changing the name would sadly mean additional disruption for consumers of the Panda library, of course): https://github.com/guardian/pan-domain-authentication/blob/053b9f85c02f907604300503f6ba5d893ea0ccc1/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala#L35

2. `settings` - the field we're reading with `panDomainSettings.settings` - is a field that can **change value over time**, because the `PanDomainAuthSettingsRefresher` will be periodically re-fetching it (and this is essential for key-rotation, as in #150): https://github.com/guardian/pan-domain-authentication/blob/053b9f85c02f907604300503f6ba5d893ea0ccc1/pan-domain-auth-core/src/main/scala/com/gu/pandomainauth/PanDomainAuthSettingsRefresher.scala#L43

...so, given those two things, it **def**initely makes sense that `AuthActions.settings` remains a `def`!

PR #165, opened 2024-11-22
Support Scala 3, drop Scala 2.12 & Play 2.8

rtyley:
### Need to import `play.api.libs.ws._` for the `BodyWritable`

Panda's `OAuth` class posts some url-encoded data (defined as `Map[String, Seq[String]]` in Scala) with `ws.url(dd.token_endpoint).post`, and this needs an implicit instance of `BodyWritable[Map[String, Seq[String]]]` in order to work! For some reason, in Scala 2, the compiler was able to find the correct implicit somewhere, but in Scala 3 we get a compilation error:

```
[error] 81 |        }.flatMap { response =>
[error]    |         ^
[error]    |Cannot find an instance of Map[K, Seq[String]] to WSBody. Define a BodyWritable[Map[K, Seq[String]]] or extend play.api.libs.ws.ahc.DefaultBodyWritables
[error]    |
[error]    |where:    K is a type variable with constraint >: String
[error]    |
[error]    |One of the following imports might fix the problem:
[error]    |
[error]    |
[error]    |One of the following imports might fix the problem:
[error]    |
[error]    |  import play.api.libs.ws.DefaultBodyWritables.writeableOf_urlEncodedForm
[error]    |  import play.api.libs.ws.WSBodyWritables.writeableOf_urlEncodedForm
[error]    |  import play.api.libs.ws.writeableOf_urlEncodedForm
```

Importing the whole `ws` package fixes the problem:

```
import play.api.libs.ws._
```

@guardian/digital-cms


pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
```diff
@@ -2,7 +2,7 @@ package com.gu.pandomainauth.service
 
 import com.gu.pandomainauth.model.{AuthenticatedUser, OAuthSettings, User}
 import play.api.libs.json.JsValue
-import play.api.libs.ws.{WSClient, WSResponse}
+import play.api.libs.ws._
```
rtyley, line 5: Panda's `OAuth` class posts some url-encoded data (defined as `Map[String, Seq[String]]` in Scala) with `ws.url(dd.token_endpoint).post`:

https://github.com/guardian/pan-domain-authentication/blob/2c378addf9cb10b9b629013bc336a6da01891e76/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala#L73-L81

This needs an implicit instance of `BodyWritable[Map[String, Seq[String]]]` in order to work! For some reason, in Scala 2, the compiler was able to find the correct implicit somewhere, but in Scala 3 we get a compilation error:

```
[error] 81 |        }.flatMap { response =>
[error]    |         ^
[error]    |Cannot find an instance of Map[K, Seq[String]] to WSBody. Define a BodyWritable[Map[K, Seq[String]]] or extend play.api.libs.ws.ahc.DefaultBodyWritables
[error]    |
[error]    |where:    K is a type variable with constraint >: String
[error]    |
[error]    |One of the following imports might fix the problem:
[error]    |
[error]    |
[error]    |One of the following imports might fix the problem:
[error]    |
[error]    |  import play.api.libs.ws.DefaultBodyWritables.writeableOf_urlEncodedForm
[error]    |  import play.api.libs.ws.WSBodyWritables.writeableOf_urlEncodedForm
[error]    |  import play.api.libs.ws.writeableOf_urlEncodedForm
```

Importing the whole `ws` package fixes the problem.

project/plugins.sbt
```diff
@@ -1,5 +1,5 @@
 // Use the Play sbt plugin for Play projects
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.9.5")
+addSbtPlugin("org.playframework" % "sbt-plugin" % "3.0.5")
```
rtyley, line 2: See:

* https://github.com/guardian/maintaining-scala-projects/issues/4

Note that the Play `sbt-plugin` is only required for the `exampleApp` app, not any of the other subprojects.

project/Dependencies.scala
```diff
@@ -49,5 +40,5 @@ object Dependencies {
 
   // provide compatibility between scala 2.12 and 2.13
   // see https://github.com/scala/scala-collection-compat/issues/208
-  val scalaCollectionCompatDependencies = Seq("org.scala-lang.modules" %% "scala-collection-compat" % "2.12.0")
+  val scalaCollectionCompat = "org.scala-lang.modules" %% "scala-collection-compat" % "2.12.0"
```
bryophyta, line 43: do we still need this dependency, given that 2.12 support is removed in this PR?

build.sbt
```diff
@@ -46,104 +46,60 @@ lazy val panDomainAuthCore = subproject("pan-domain-auth-core")
...
-  panDomainAuthHmac_2_8,
   panDomainAuthHmac_2_9,
   panDomainAuthHmac_3_0,
   exampleApp
-).settings(sonatypeReleaseSettings)
- .settings(
-  organization := "com.gu",
+).settings(
   publish / skip := true,
+  // releaseVersion := ReleaseVersion.fromAggregatedAssessedCompatibilityWithLatestRelease().value,
```
bryophyta, line 92: should this line be commented out?

rtyley: Unfortunately yes, due to:

https://github.com/guardian/gha-scala-library-release-workflow/issues/33

project/Dependencies.scala
rtyley: Good spot! Done with 2b1a114!

bryophyta (approved):
My comments have been addressed, and we discussed off-thread too. Looks good to me!

PR #167, opened 2024-12-06
node: Correct the time comparison used for cookie expiry

andrew-nowak:
@guardian/digital-cms

`verifyUser` expects a full unix timestamp, but `Date.getMilliseconds()` will only return a value between 0 and 999 (the millisecond component of the timestamp). This should be `.getDate()` instead! Otherwise the current time always will appear to be within the first second of 1970, which will always be less than the expiry time encoded in the panda cookie.


twrichards (approved):
paired on this but LGTM

pan-domain-node/src/panda.ts
```diff
@@ -57,6 +57,8 @@ export function verifyUser(pandaCookie: string | undefined, publicKey: string, c
         return { status: AuthenticationStatus.INVALID_COOKIE };
     }
 
+    const currentTimestamp = currentTime.getTime();
```
joelochlann, line 60: Could you add seconds/milliseconds to the variable name to help reduce the kind of confusion that led to this mistake?

joelochlann:
```suggestion
    const currentTimestampInMilliseconds = currentTime.getTime();
```

pan-domain-node/test/panda.test.ts
```diff
@@ -13,28 +13,29 @@ import {decodeBase64} from "../src/utils";
...
         const testCookie = data + ".1234";
 
-        expect(verifyUser(testCookie, publicKey, 0, guardianValidation).status).toBe(AuthenticationStatus.INVALID_COOKIE);
+        expect(verifyUser(testCookie, publicKey, new Date(0), guardianValidation).status).toBe(AuthenticationStatus.INVALID_COOKIE);
     });
 
     test("return expired", () => {
-        const someTimeInTheFuture = 5678;
+        const someTimeInTheFuture = new Date(5678);
+        expect(someTimeInTheFuture.getTime()).toBe(5678);
```
joelochlann, line 28: Why did this test not catch the problem originally? Are there any extra tests or comments that could be added to help avoid or clarify that pitfall?

pan-domain-node/src/panda.ts
```diff
@@ -108,7 +110,7 @@ export class PanDomainAuthentication {
     getPublicKey(): Promise<string> {
         return this.publicKey.then(({ key, lastUpdated }) => {
             const now = new Date();
-            const diff = now.getMilliseconds() - lastUpdated.getMilliseconds();
+            const diff = now.getTime() - lastUpdated.getTime();
```
joelochlann, line 113: Looks like this is a fix for separate problem? Could we add some tests for this?

pan-domain-node/src/panda.ts
joelochlann: (Oh, although I know see that the problem is not to do with seconds vs milliseconds, it's about getting an epoch time vs the milliseconds since the last second. Still, can't hurt to be explicit)

pan-domain-node/test/panda.test.ts
joelochlann: Oh! It's because it doesn't actually test the `verify` function which is what's called by the consumer of this library 😂🤦🏻‍♂️

Let's **definitely** add a tests for the actual public API of the library:
- `stop()`
- `getPublicKey()`
- `verify()`

andrew-nowak: several hours of fighting with jest's mocking mechanisms later, I've got a handful of tests added now

joelochlann: You're an absolute hero! 🫡 Yeah it occurred to me after writing this comment that the annoyance of spoofing `new Date()` was probably what caused someone to avoid writing tests for the public API in the first place!

rtyley:
See https://chat.google.com/room/AAAAUqoLvh4/K6HTgS4B8QI/0UxdWnNQUsM?cls=10 for some additional context around the issues this PR fixes.

PR #172, opened 2024-12-09
changesets action won't run on node 12, so upgrade to 20

andrew-nowak:
@guardian/digital-cms

changesets action won't run on node 12, so upgrade to 20


twrichards (approved):
Good thing anyway

PR #179, opened 2025-01-15
pass the user data with multifactor checked to invalidUserMessage

andrew-nowak:
Currently this function passes the version of the user data before the multifactor status has been checked into `invalidUserMessage`, which means that if the function has been overridden to provide specific messages about the user's invalid status, the user may see warnings about MFA not being active, despite that MFA is in fact active.

Update to pass in the latest version of the user data, which matches the version passed to the validity checker


bryophyta (approved):
looks good!

PR #182, opened 2025-02-19
key-rotation: monitor non-active-key use

rtyley:
As we perform the key-rotation enabled by #150, it'll be good to have insight into how many users are still using cookies signed by old keys (to be precise, keys that are not the _active_ key) - we should see it tail off within the space of 1 hour. 

This change adds monitoring that records how often cookies signed by non-active keys are being validated - when a non-active key is used, the stats are logged out.

> NON_ACTIVE_KEY_USED (used='1xpqc', active='6k065'): active_freq=Map(false -> 22, true -> 3) key_freq=Map('1xpqc' -> 25)

### Guava, rate-limiting & logging

The frequency of logging is throttled by Guava's [`RateLimiter`](https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/RateLimiter.java) to once every 10 seconds per JVM. Consequently, **this update adds [Guava](https://github.com/google/guava) as a dependency**.

## Testing

I've tried this PR out using a preview release on [tagmanager.code.dev-gutools.co.uk](https://tagmanager.code.dev-gutools.co.uk/), where it worked as intended:

* https://github.com/guardian/tagmanager/pull/547

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -87,7 +87,7 @@ lazy val root = Project("pan-domain-auth-root", file(".")).aggregate(
   exampleApp
 ).settings(
   publish / skip := true,
-  // releaseVersion := ReleaseVersion.fromAggregatedAssessedCompatibilityWithLatestRelease().value,
+  releaseVersion := ReleaseVersion.fromAggregatedAssessedCompatibilityWithLatestRelease().value,
   releaseCrossBuild := true, // true if you cross-build the project for multiple Scala versions
   releaseProcess := Seq[ReleaseStep](
     checkSnapshotDependencies,
--- a/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CryptoConf.scala
+++ b/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CryptoConf.scala
@@ -1,7 +1,9 @@
 package com.gu.pandomainauth.service
 
 import com.gu.pandomainauth.Settings._
+import com.gu.pandomainauth.internal.{KeyHashId, NonActiveKeyMonitoring}
 import com.gu.pandomainauth.service.Crypto.keyFactory
+import com.gu.pandomainauth.service.CryptoConf.Change.ActiveKey.TransitionCriteria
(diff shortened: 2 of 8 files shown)


pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CryptoConf.scala
```diff
@@ -23,9 +24,22 @@ object CryptoConf {
...
+
     private[CryptoConf] def acceptsActiveKeyFrom(other: Verification): Boolean = acceptedPublicKeys.contains(other.activePublicKey)
 
-    def decode[A](f: PublicKey => Option[A]): Option[A] = acceptedPublicKeys.flatMap(f(_)).headOption
+    def decode[A](f: PublicKey => Option[A]): Option[A] = {
+      (for {
+        (key, keyId) <- acceptedKeysWithIds
+        result <- f(key)
+      } yield {
+        NonActiveKeyMonitoring.instance.monitor(keyId, activeKeyId)
```
rtyley, line 39: This is the essential change in this PR - we call out to a logging class, passing enough information for it to record key usage, and record if the key is active - or not.

bryophyta (approved):
The code looks good to me, and we've discussed off-thread. There are some new abstractions but they seem well motivated for getting the logging behaviour right, and they all seem understandable to me.

build.sbt
```diff
@@ -87,7 +87,7 @@ lazy val root = Project("pan-domain-auth-root", file(".")).aggregate(
   exampleApp
 ).settings(
   publish / skip := true,
-  // releaseVersion := ReleaseVersion.fromAggregatedAssessedCompatibilityWithLatestRelease().value,
+  releaseVersion := ReleaseVersion.fromAggregatedAssessedCompatibilityWithLatestRelease().value,
```
rtyley, line 90: Re-enabling this after we had to disable it for https://github.com/guardian/pan-domain-authentication/pull/165 - the new Scala 3 artifacts are available:

* https://github.com/guardian/gha-scala-library-release-workflow/issues/33

PR #184, opened 2025-03-12
Adapt `key-rotation` script for Scala 3, stop it from building under Scala 2

bryophyta:
## Problem

The Scala script for generating [new key-rotation config](https://github.com/guardian/pan-domain-authentication/pull/150) was broken by the addition of Scala 3 support in https://github.com/guardian/pan-domain-authentication/pull/165.

This was because the script used the preferred Scala 2 syntax for creating CLI apps: the `App` trait which (somewhat magically) provides an `args` value using the `DelayedInit` trait. Scala 3 removed support for that trait, meaning that `args` was no longer populated - giving not a compile-time error, but a cryptic runtime error:

```
Exception in thread "sbt-bg-threads-1" java.lang.ExceptionInInitializerError
        at com.gu.pandomainauth.CryptoConfForRotation.main(CryptoConfForRotation.scala)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at sbt.Run.invokeMain(Run.scala:144)
        at sbt.Run.execute$1(Run.scala:94)
        at sbt.Run.$anonfun$runWithLoader$5(Run.scala:121)
        at sbt.Run$.executeSuccess(Run.scala:187)
        at sbt.Run.runWithLoader(Run.scala:121)
        at sbt.Defaults$.$anonfun$bgRunTask$6(Defaults.scala:2030)
        at sbt.Defaults$.$anonfun$termWrapper$2(Defaults.scala:1969)
        at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
        at scala.util.Try$.apply(Try.scala:213)
        at sbt.internal.BackgroundThreadPool$BackgroundRunnable.run(DefaultBackgroundJobService.scala:367)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.NullPointerException
        at java.base/java.lang.reflect.Array.getLength(Native Method)
        at scala.collection.ArrayOps$.headOption$extension(ArrayOps.scala:219)
        at com.gu.pandomainauth.CryptoConfForRotation$.<clinit>(CryptoConfForRotation.scala:30)
        ... 16 more
```

We want the script to run under Scala 3, and use the new Scala 3 syntax for CLI apps (ie the `@main` method annotation, and [args-from-method-parameters](https://docs.scala-lang.org/scala3/book/methods-main-methods.html#command-line-arguments)), but leaving the script where it was (inside the tests of `pan-domain-auth-verification`) meant that when CI ran `++test`, `sbt` would attempt to compile the script under both Scala 2 & 3, which would fail on Scala 2 where obviously the Scala 3 syntax wasn't supported.

## Fix

There does exist [syntax that works for both Scala 2 & Scala 3](https://docs.scala-lang.org/scala3/book/methods-main-methods.html#backwards-compatibility-with-scala-2), but really, it probably makes sense to move the script out to its own subproject, where it will only be compiled for Scala 3 and additionally can be invoked from the sbt console much more concisely:

#### Before

```
pan-domain-auth-verification / Test / run panda.settings
```

#### After

```
key-rotation/run panda.settings
```

I've also tried to tidy up the sbt build script a bit too!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,66 +7,60 @@ import play.sbt.PlayImport.PlayKeys.*
 import sbtversionpolicy.withsbtrelease.ReleaseVersion
 
 ThisBuild / scalaVersion := "3.3.5"
-ThisBuild / crossScalaVersions := Seq(
-  scalaVersion.value,
-  "2.13.16"
-)
+val crossCompileScalaVersions = crossScalaVersions := Seq(scalaVersion.value, "2.13.16")
 
 val commonSettings = Seq(
-  organization := "com.gu",
-  licenses := Seq(License.Apache2),
   Test / fork := false,
-  scalacOptions := Seq(
-    "-feature",
-    "-deprecation",
-    "-release:11"
-  ),
   libraryDependencies ++= testDependencies,
   Test / testOptions +=
     Tests.Argument(TestFrameworks.ScalaTest, "-u", s"test-results/scala-${scalaVersion.value}", "-o")
 )
 
-def subproject(path: String): Project =
-  Project(path, file(path)).settings(commonSettings: _*)
-
-lazy val panDomainAuthVerification = subproject("pan-domain-auth-verification")
-  .settings(
-    libraryDependencies
-      ++= cryptoDependencies
-      ++ awsDependencies
-      ++ testDependencies
-      ++ loggingDependencies,
-  )
-
-lazy val panDomainAuthCore = subproject("pan-domain-auth-core")
-  .dependsOn(panDomainAuthVerification)
-  .settings(
-    libraryDependencies
-      ++= awsDependencies
-      ++ googleDirectoryApiDependencies
-      ++ cryptoDependencies
-      ++ testDependencies,
-  )
-
-def playBasedProject(playVersion: PlayVersion, projectPrefix: String, srcFolder: String) =
-  subproject(s"$projectPrefix-${playVersion.suffix}").settings(
-    sourceDirectory := (ThisBuild / baseDirectory).value / srcFolder / "src"
-  )
-
-def playSupportFor(playVersion: PlayVersion) =
-  playBasedProject(playVersion, "pan-domain-auth", "pan-domain-auth-play").settings(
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -7,82 +7,83 @@ import play.sbt.PlayImport.PlayKeys.*
 import sbtversionpolicy.withsbtrelease.ReleaseVersion
 
 ThisBuild / scalaVersion := "3.3.5"
-ThisBuild / crossScalaVersions := Seq(
-  scalaVersion.value,
-  "2.13.16"
-)
```
rtyley, line 13: This setting comes out, because we no longer want _all_ projects to be cross-compiled with both Scala 2 & Scala 3 - the key rotation script is just for Scala 3.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/Settings.scala
```diff
@@ -32,7 +32,7 @@ trait FailureWithCause extends SettingsFailure {
 }
 
 case class SettingsDownloadFailure(cause: Throwable) extends FailureWithCause {
-  override val description: String = "Unable to download public key"
+  override val description: String = "Unable to access settings file"
```
rtyley, line 35: This change gives a less misleading error message when the user runs the script supplying a path that doesn't point to a file:

```
sbt key-rotation/run /tmp/file-does-not-exist.settings
Loading current conf from /tmp/file-does-not-exist.settings
Unable to access settings file
```

dblatcher (approved):
This looks good to me - I don't think we have a way to dry-run or test the script locally.

PR #186, opened 2025-03-14
Extract non play hmac artifact

rtyley:


build.sbt
```diff
@@ -82,12 +88,13 @@ lazy val root = Project("pan-domain-auth-root", file(".")).aggregate(
...
   panDomainAuthPlay_2_9,
   panDomainAuthPlay_3_0,
+  panDomainAuthHmac,
   panDomainAuthHmac_2_9,
   panDomainAuthHmac_3_0,
   exampleApp
 ).settings(
   publish / skip := true,
-  releaseVersion := ReleaseVersion.fromAggregatedAssessedCompatibilityWithLatestRelease().value,
+  // releaseVersion := ReleaseVersion.fromAggregatedAssessedCompatibilityWithLatestRelease().value,
```
jonathonherbert, line 97: Still needed?

hmac/core/src/main/scala/com/gu/pandahmac/HMACClient.scala
```diff
@@ -0,0 +1,21 @@
+package com.gu.pandahmac
+
+import com.gu.hmac.HMACHeaders
+
+import java.net.URI
+
+/**
```
jonathonherbert, line 7: This comment might not be true over time, and IIUC as a result of this release should no longer be true soon-ish: hostage to fortune, or pragmatically helpful in what is usually a slow-moving domain?

hmac/play/play_2-9/src/test/scala/com/gu/pandahmac/HmacSecretsTest.scala
```diff
@@ -0,0 +1,60 @@
+package com.gu.pandahmac
+
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+import java.net.URI
+import java.time.format.DateTimeFormatter
+import java.time.{Clock, Instant, ZoneId}
+
+class HMACHeadersTest extends AnyWordSpec with Matchers {
```
jonathonherbert, line 10: Ooh, new in this PR? 👏

jonathonherbert (approved):
A useful addition, and a lovely spruce. Thanks for this, @rtyley!

Tested locally in `flexible-content` and this looks like a drop-in replacement as the classpath remains the same: that's great! (But surfacing here in case that's not expected.)

build.sbt
rtyley: Unfortunately we have to comment the line out until after the release due to this issue:

* https://github.com/guardian/gha-scala-library-release-workflow/issues/33

jonathonherbert: Crikey! Thanks.

hmac/core/src/main/scala/com/gu/pandahmac/HMACClient.scala
rtyley: This particular class actually isn't used anywhere yet, so I guess it's already misleading! https://github.com/guardian/flexible-content/pull/5296 will start consuming it, and then it makes sense to role it out to `workflow`, `typerighter`, etc.

I think you're right that it's the kind of comment that can go out-of-date over time, I'll delete it...

hmac/play/play_2-9/src/test/scala/com/gu/pandahmac/HmacSecretsTest.scala
rtyley: > Ooh, new in this PR?

Ah, I'm afraid not - I accidentally made two copies of the same test file! I've removed the second copy with 9af8d90.

PR #177, opened 2024-12-10
Make an expired cookie

rtyley:

PR #187, opened 2025-03-19
Strongly-type time fields

rtyley:
This changes extracts part of https://github.com/guardian/pan-domain-authentication/pull/185: it takes the parts that strongly-type the time fields to use the `java.time` classes `Instant` & `Duration`, while leaving the grace-period-related parts for that other PR.

Note that some consuming code in [several apps](https://github.com/search?q=org%3Aguardian+%22apiGracePeriod+%3D%22+NOT+repo%3Aguardian%2Fpan-domain-authentication&type=code) will need to be updated to reflect the new `Instant` type of `apiGracePeriod`.

### The time-unit of the `expires` field... _milliseconds_ or _seconds_ since the epoch?

`expires` is a field in two places:

* the serialised, Base64-encoded value in Panda's `gutoolsAuth-assym` cookie
* the field on the [`AuthenticatedUser`](https://github.com/guardian/pan-domain-authentication/blob/fb75c04bb7436faae1f71f93ebd7532a39ea8a51/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/model/AuthenticatedUser.scala#L5) case class - where it was represented by a `Long`

It's actually in **milliseconds**, but it was treated [inconsistently](https://github.com/guardian/pan-domain-authentication/pull/187/files#r2003779712) as _seconds_ in Panda's test suite.

We can be certain that `expires` is in milliseconds for these two reasons:

1. Examining a real `gutoolsAuth-assym` cookie and Base64-decoding it, we get a payload like this:
    > firstName=Roberto&lastName=Tyley&...&**expires=1742386198000**&multifactor=true
    
    The value `expires=1742386198000` only [makes sense](https://www.epochconverter.com/) if it's in _milliseconds_ :
    image
1.  The current code for determining `isExpired` compares `expired` to [`new Date().getTime`](https://docs.oracle.com/javase/8/docs/api/java/util/Date.html#getTime--):
    https://github.com/guardian/pan-domain-authentication/blob/3b00ea4d92a95b8d4eda718cf01d1c3a1fa696b5/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/model/AuthenticatedUser.scala#L7
    `new Date().getTime` returns _"the number of **milliseconds** since January 1, 1970, 00:00:00 GMT"_

## See also

* https://github.com/guardian/pan-domain-authentication/pull/188

### DIFF
--- a/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
+++ b/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
@@ -8,6 +8,7 @@ import play.api.mvc.{RequestHeader, Result}
 
 import java.math.BigInteger
 import java.security.SecureRandom
+import java.time.Instant
 import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.{ExecutionContext, Future}
 import scala.language.postfixOps
@@ -96,8 +97,8 @@ class OAuth(config: OAuthSettings, system: String, redirectUrl: String)(implicit
                   ),
                   authenticatingSystem = system,
                   authenticatedIn = Set(system),
-                  jwt.claims.exp * 1000,
-                  false
+                  expires = Instant.ofEpochSecond(jwt.claims.exp), // https://stackoverflow.com/a/39926886/438886
+                  multiFactor = false
                 )
               }
             }
--- a/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
+++ b/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
@@ -97,7 +97,11 @@ class OAuth(config: OAuthSettings, system: String, redirectUrl: String)(implicit
                   ),
                   authenticatingSystem = system,
                   authenticatedIn = Set(system),
-                  expires = Instant.ofEpochSecond(jwt.claims.exp), // https://stackoverflow.com/a/39926886/438886
+                  // The JWT standard specifies that `exp` is a `NumericDate`,
+                  // which is defined as an epoch time in *seconds*
+                  // (unlike the Panda cookie `expires` which is in milliseconds)
+                  // https://www.rfc-editor.org/rfc/rfc7519#section-4.1.4
+                  expires = Instant.ofEpochSecond(jwt.claims.exp),
                   multiFactor = false
                 )
               }
--- a/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
+++ b/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
@@ -4,6 +4,7 @@ import com.gu.pandomainauth.model.{AuthenticatedUser, User}
 import com.gu.pandomainauth.service.CookieUtils.CookieIntegrityFailure.{MalformedCookieText, MissingOrMalformedUserData, SignatureNotValid}
 import com.gu.pandomainauth.service.CryptoConf.{Signing, Verification}
 
+import java.time.Instant
 import scala.util.Try
 
 object CookieUtils {
@@ -23,7 +24,7 @@ object CookieUtils {
       authUser.user.avatarUrl.map(a => s"&avatarUrl=$a").getOrElse("") +
       s"&system=${authUser.authenticatingSystem}" +
       s"&authedIn=${authUser.authenticatedIn.mkString(",")}" +
-      s"&expires=${authUser.expires}" +
+      s"&expires=${authUser.expires.toEpochMilli}" +
       s"&multifactor=${authUser.multiFactor}"
 
   private[service] def deserializeAuthenticatedUser(serializedForm: String): Option[AuthenticatedUser] = {
@@ -45,7 +46,7 @@ object CookieUtils {
       user = User(firstName, lastName, email, data.get("avatarUrl")),
       authenticatingSystem = system,
       authenticatedIn = Set(authedIn.split(",").toSeq :_*),
-      expires = expires,
+      expires = Instant.ofEpochMilli(expires),
       multiFactor = multiFactor
     )
   }
--- a/pan-domain-auth-verification/src/test/scala/com/gu/pandomainauth/PanDomainTest.scala
+++ b/pan-domain-auth-verification/src/test/scala/com/gu/pandomainauth/PanDomainTest.scala
@@ -7,7 +7,10 @@ import org.scalatest.Inside
 import org.scalatest.freespec.AnyFreeSpec
 import org.scalatest.matchers.should.Matchers
 
-import java.util.Date
+import java.time.Duration.{ZERO, ofMillis}
+import java.time.Instant.now
+import java.time.temporal.ChronoUnit.MILLIS
+import java.time.{Duration, Instant}
 
 class PanDomainTest extends AnyFreeSpec with Matchers with Inside {
   import com.gu.pandomainauth.service.TestKeys._
@@ -16,14 +19,22 @@ class PanDomainTest extends AnyFreeSpec with Matchers with Inside {
   def authStatus(
     cookieData: String,
     validateUser: AuthenticatedUser => Boolean = _ => true,
-    apiGracePeriod: Long = 0,
+    apiGracePeriod: Duration = ZERO,
     system: String = "testsuite",
     cacheValidation: Boolean = false,
     forceExpiry: Boolean = false,
   ) = PanDomain.authStatus(cookieData, OnlyVerification(testPublicKey.key), validateUser, apiGracePeriod, system, cacheValidation, forceExpiry)
 
   "authStatus" - {
-    val authUser = AuthenticatedUser(User("test", "user", "test.user@example.com", None), "testsuite", Set("testsuite"), new Date().getTime + 86400, multiFactor = true)
+    val authUser = AuthenticatedUser(
+      User("test", "user", "test.user@example.com", None),
+      "testsuite",
+      Set("testsuite"),
+      // The expiry is serialised to millisecond accuracy
+      // so this needs to be at the same precision for comparison.
+      now().plusMillis(86400).truncatedTo(MILLIS),
+      multiFactor = true
+    )
     val validCookieData = CookieUtils.generateCookieData(authUser, signingWith(testPrivateKey.key))
 
     "returns `Authenticated` for valid cookie data that passes the validation check" in {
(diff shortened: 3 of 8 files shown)


pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
```diff
@@ -96,8 +97,8 @@ class OAuth(config: OAuthSettings, system: String, redirectUrl: String)(implicit
                   ),
                   authenticatingSystem = system,
                   authenticatedIn = Set(system),
-                  jwt.claims.exp * 1000,
-                  false
+                  expires = Instant.ofEpochSecond(jwt.claims.exp), // https://stackoverflow.com/a/39926886/438886
+                  multiFactor = false
```
rtyley, line 101: As explained in this [Stackoverflow answer](https://stackoverflow.com/a/39926886/438886), RFC 7519 [states that](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.4) `exp` is a `NumericDate`, and that `NumericDate` is the number of **seconds** (not milliseconds) since the epoch - which is why we were multiplying by `1000` to get to the number of milliseconds for `expires` in the old code.

Now that `expires` is a strongly-typed `Instant`, we can use the explicit `Instant.ofEpochSecond()` method to make clear what conversion we're using.

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
```diff
@@ -23,7 +24,7 @@ object CookieUtils {
       authUser.user.avatarUrl.map(a => s"&avatarUrl=$a").getOrElse("") +
       s"&system=${authUser.authenticatingSystem}" +
       s"&authedIn=${authUser.authenticatedIn.mkString(",")}" +
-      s"&expires=${authUser.expires}" +
+      s"&expires=${authUser.expires.toEpochMilli}" +
```
rtyley, line 27: As we've established, the `expires` field in the cookie payload is in **milliseconds**:

> firstName=Roberto&lastName=Tyley&...&**expires=1742386198000**&multifactor=true

pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala
```diff
@@ -45,7 +46,7 @@ object CookieUtils {
       user = User(firstName, lastName, email, data.get("avatarUrl")),
       authenticatingSystem = system,
       authenticatedIn = Set(authedIn.split(",").toSeq :_*),
-      expires = expires,
+      expires = Instant.ofEpochMilli(expires),
```
rtyley, line 49: Here we're reading _out_ of the cookie payload, and `expires` field in the Panda cookie payload is *milliseconds**.

pan-domain-auth-verification/src/test/scala/com/gu/pandomainauth/PanDomainTest.scala
```diff
@@ -49,30 +60,48 @@ class PanDomainTest extends AnyFreeSpec with Matchers with Inside {
...
+      // Therefore, we have one minute left of the custom 2-hour grace period.
+      val expiredAuthUser = authUser.copy(expires = now() minus customGracePeriod.minusMinutes(1))
       val cookieData = CookieUtils.generateCookieData(expiredAuthUser, signingWith(testPrivateKey.key))
 
-      authStatus(cookieData, apiGracePeriod = 3600) shouldBe a [GracePeriod]
+      authStatus(cookieData, apiGracePeriod = Some(customGracePeriod)) shouldBe a [GracePeriod]
+    }
+
+
+    "returns `Expired` if cookie has not expired, but forceExpiry is set" in {
```
joelochlann, line 102: @rtyley this is the only mention of `forceExpiry` in the tests.

It makes it clear that when the cookie is fresh, setting `forceExpiry` to `true` will give an `AuthenticationStatus` of `Expired`.

However, it doesn't specify the behaviour wrt grace period.

It looks to me like the existing behaviour in `checkStatus` is that if you are in period C below (aka the grace period), then `forceExpiry` will never work!

```
Panda cookie:  issued       expires
               |-- 1 hour --|--24 hours------------------------------|
            A--|----B-------|--C-------------------------------------|--D----
```


This seems accidental. Why should the grace period "protect" you from forced expiry?

joelochlann: One other useful thing I discovered: the backstory of `forceExpiry`, apparently just for debug purposes https://github.com/guardian/pan-domain-authentication/blob/a9e2027516ad1f7e424e7f2dd8b2c918c1482e76/pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/action/Actions.scala#L105-L109

rtyley: > the backstory of `forceExpiry`, apparently just for debug purposes

Ah yeah, just found the PR:

* https://github.com/guardian/pan-domain-authentication/pull/117

...it's a convenient way to simulate an expired cookie, but I don't think it addresses the distinction between expired/grace-period/acceptable/fresh... what happens if you want to be in- or out - of the grace-period?

Possibly finishing off this debug tool might be an alternative - not as convenient to use though:

* https://github.com/guardian/pan-domain-authentication/pull/177

...but at least it would simplify the code in `checkStatus()`.

pan-domain-auth-verification/src/test/scala/com/gu/pandomainauth/PanDomainTest.scala
```diff
@@ -16,14 +19,22 @@ class PanDomainTest extends AnyFreeSpec with Matchers with Inside {
...
     validateUser: AuthenticatedUser => Boolean = _ => true,
-    apiGracePeriod: Long = 0,
+    apiGracePeriod: Duration = ZERO,
     system: String = "testsuite",
     cacheValidation: Boolean = false,
     forceExpiry: Boolean = false,
   ) = PanDomain.authStatus(cookieData, OnlyVerification(testPublicKey.key), validateUser, apiGracePeriod, system, cacheValidation, forceExpiry)
 
   "authStatus" - {
-    val authUser = AuthenticatedUser(User("test", "user", "test.user@example.com", None), "testsuite", Set("testsuite"), new Date().getTime + 86400, multiFactor = true)
```
rtyley, line 26: Note that [`new Date().getTime`](https://docs.oracle.com/javase/8/docs/api/java/util/Date.html#getTime--) returns a value in _milliseconds_.

`86400` milliseconds is 86.4 seconds, while `86400` seconds is 24 hours, so this strange choice of `86400` probably arose from confusion between seconds and milliseconds - we've chosen not to correct this to a more intentional value here, to demonstrate that the tests still pass with minimal alteration.

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
```diff
@@ -96,8 +97,8 @@ class OAuth(config: OAuthSettings, system: String, redirectUrl: String)(implicit
                   ),
                   authenticatingSystem = system,
                   authenticatedIn = Set(system),
-                  jwt.claims.exp * 1000,
-                  false
+                  expires = Instant.ofEpochSecond(jwt.claims.exp), // https://stackoverflow.com/a/39926886/438886
```
joelochlann, line 100: On reading this through with fresh eyes, I actually think the seconds/milliseconds stuff is confusing enough that it warrants a more explicit comment.

My first question when seeing `ofEpochSecond` would be "wait, was this a mistake or intentional?", and I don't think linking to StackOverflow is enough to reassure the reader that the JWT/Panda distinction has been deliberately considered.

```suggestion
                  // The JWT standard specifies that `exp` is a `NumericDate`,
                  // which is defined as an epoch time in *seconds*
                  // (unlike the Panda cookie `expires` which is in milliseconds)
                  // https://www.rfc-editor.org/rfc/rfc7519#section-4.1.4
                  expires = Instant.ofEpochSecond(jwt.claims.exp),
```

pan-domain-auth-verification/src/test/scala/com/gu/pandomainauth/PanDomainTest.scala
emdash-ie: Now that the tests have been demonstrated to pass (https://github.com/guardian/pan-domain-authentication/actions/runs/13951682077/job/39052570386), can we change it to a more sensible value (presumably `.plusSeconds(86400)`?) by adding an extra commit to the PR?

rtyley: We do have that change added into a subsequent PR (https://github.com/guardian/pan-domain-authentication/pull/185) but I can pull that change onto this one 👍

pan-domain-auth-play/src/main/scala/com/gu/pandomainauth/service/OAuth.scala
rtyley: > I don't think linking to StackOverflow is enough to reassure the reader that the JWT/Panda distinction has been deliberately considered.

Ah, fair enough! You're right that there could always be doubt in someone's mind that converting from seconds is correct, when [elsewhere](https://github.com/guardian/pan-domain-authentication/blob/f03046ae00f9f60ae2b9b6526ccff68452161f2d/pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/service/CookieUtils.scala#L49) we've converted from milliseconds.

Personally tho, I avoid inline comments _within_ code blocks (I'm a _bit_ friendlier to comments against a class, method or field). They're prone to:

* **getting separated from the code they relate to** - many times in old codebases, I’ve seen lines of code deleted, _without_ their explanatory comment on the line above being deleted as well - leaving a confusing comment, apparently speaking to a different line of code further down.
* **becoming outdated** - they speak with an authority and _apparent_ currentness there's no reasonable mechanism for enforcing: code comments _look_ like they apply to the code _as it is right now_. In contrast documentation we write in PR descriptions/comments is understood in its context - when others come back to read it (which admittedly, assumes they've done a `git blame` and found the PR!), they understand the _age_ of that context: they understand that it was true when the PR was written, but might _not_ be now.

One complementary approach might be having a test that actually _exercises_ this code (we would probably want to refactor the construction of the `AuthenticatedUser` out into its own method for easier unit testing, and supply reference JSON for both the token and OAuth response) - we could explicitly assert that `expires` ends up taking a sensible value. If someone thinks `.ofEpochSecond(jwt.claims.exp)` is wrong, and tries to change it to `ofEpochMillis`, the test will fail, and there's no way they can make it pass without either editing the reference json or changing the expected `expires` value to a ridiculous value (thousands of years in the future!).

Incidentally, note that in the new comment, there's a statement that's vulnerable to becoming untrue over time:

```
// (unlike the Panda cookie `expires` which is in milliseconds)
```

...the format of the Panda cookie could change in future - it could even become a JWT, and be in seconds! If that were to happen, there's no test that would fail and say "Hey, there's now this comment in our codebase which is confusing/misleading!" - the comment would have to be manually spotted, and corrected.

Having said all that, I guess I just wanted to express my [personal](https://docs.google.com/document/d/19Ph05_rACj-EiX3GUCHuErAR_iy4kXN-NptOd8IAkDk/edit?usp=sharing) preferences on code documentation, and there are certainly arguments against them too. Please forgive me for banging on - and thank you for adding the clarification!

pan-domain-auth-verification/src/test/scala/com/gu/pandomainauth/PanDomainTest.scala
joelochlann: So, I **think** this weirdness is fixed by the refactor in https://github.com/guardian/pan-domain-authentication/pull/188!

But we should add a test to that PR to check. Let's get this one merged first.

joelochlann:
> This looks good to me, pending the resolution of the conversation about forceExpiry and grace periods.

I've resolved that and added a todo to #188, to ensure we add a test around that

PR #188, opened 2025-03-19
Introduce fresh acceptable and refactor check status

rtyley:


pan-domain-auth-verification/src/main/scala/com/gu/pandomainauth/PanDomain.scala
```diff
@@ -21,32 +21,12 @@ object PanDomain {
...
-    } else {
-      // user has not expired but has failed validation checks
-      NotAuthorized(authedUser)
-    }
+    val cookieAge = authedUser.cookieAge
+    
+    if (!cookieAge.isAcceptable(apiGracePeriod) || forceExpiry) Expired(authedUser)
+    else if ((cacheValidation && authedUser.authenticatedIn(system)) || validateUser(authedUser)) {
+      if (cookieAge.isFresh()) Authenticated(authedUser) else GracePeriod(authedUser)
+    } else NotAuthorized(authedUser)
```
rtyley, line 29: @joelochlann & myself have walked through this replacement code and have satisfied ourselves that it's equivalent to the old code, but clearer and shorter.

rtyley:
> That might want to be qualified a bit, as we only expect it to be accepted by API endpoints. HTML endpoints should in fact redirect to auth, if they get a cookie that is acceptable but not fresh.

Ah yeah, good point...!

> Also, that behaviour is at the discretion of the calling code, and is only enforced if you used the Play-specific abstractions in `Actions.scala`, which not everyone does (e.g. [s3 uploader](https://github.com/guardian/s3-upload/blob/a976a57c3804f65327a9de35e37b8e4cadc09d9c/app/lib/PanAuth.scala#L31-L39)).
> 
> Otherwise, if you just use `PanDomain.authStatus` directly, it's up to you how you treat `Authenticated` vs `GracePeriod`.

I think we still want to aim for consistency across all those services, right?

> I suppose hiding them would force a more significant Composer refactor when upgrading the library. But that might not be a bad thing, given that Composer's custom logic was the source of the 10-year security hole that we [recently patched](https://github.com/guardian/flexible-content/pull/5096).

Hmmm - maybe we should be aiming to bring Composer into the fold, and reduce/remove its custom logic - and then, I guess, nothing external to Panda would be referencing the fresh/acceptable fields.

joelochlann:
> I think we still want to aim for consistency across all those services, right?

Yeah absolutely. The Play integration is probably the only way to truly enforce it. But I think hiding these methods and having `AuthenticationStatus` as the only public API would help, I think.

It would also make it more consistent with the node library, which doesn't expose any such methods: https://github.com/guardian/pan-domain-node/blob/main/src/api.ts

PR #189, opened 2025-03-19
Update GitHub codeowners to the workflow-and-collaboration team

rtyley:
It looks like @guardian/workflow-and-collaboration are the right team for ownership of the pan-domain-authentication library? As well as authentication feeling like quite a key thing to both workflow and collaboration, the Panda service depends on the Panda settings files, which live in an S3 bucket in the AWS Workflow account!

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @guardian/digital-cms
+* @guardian/workflow-and-collaboration


emdash-ie (approved):
We had a chat about this and we’re totally happy to take on ownership for pan-domain-authentication. We should make sure we have admin rights on the repository as well as being listed as code owners, and we should take the same ownership of https://github.com/guardian/login.gutools/. (I believe @MaxWalker2010 has updated the ownership list in the relevant trello board to include panda as owned by us.)

emdash-ie:
@rtyley Do you have the necessary access to give our team admin rights on this repo and on login.gutools?

rtyley:
> @rtyley Do you have the necessary access to give our team admin rights on this repo and on login.gutools?

Sure, I've granted `Admin` to @guardian/workflow-and-collaboration on these repos:

- https://github.com/guardian/pan-domain-authentication/settings/access
- https://github.com/guardian/pan-domain-node/settings/access
- https://github.com/guardian/login.gutools/settings/access

I've also given `Write` access on:

- https://github.com/guardian/data-science-recipes/settings/access (because it contains an implementation of Panda [in Python](https://github.com/guardian/data-science-recipes/blob/9f9be9ffbb8473e6b6b541d35c654427bd676409/src/server/panda.py#L85-L89), which needs to be considered when [changing things](https://github.com/guardian/data-science-recipes/issues/119) about Panda)

